PR #8, opened 2017-09-25
Add config from CSX-1

stuwilkins:
Attn @ambarb @wen-hu @cmazzoli 

This is the initial cut from the profile directory into this structure. I would like to review all that is in here and see what can be replaced / refactored and what is obsolete.


ambarb:
@stuwilkins I a little confused. Why are the files in `../startup` no longer numbered?  Are we going to hope alphabetical order is enough to always keep things running in the correct order?

ambarb:
I also cannot find the tardis setup.  Did you leave that out on purpose or can I simply not find it?  I've looked in ../ophyd/devices.py and /startup/endstation.py.

stuwilkins:
@ambarb 

Thanks! I was thinking that we will define in `__init__.py`

```python
from .detectors import *
from .accelerator import *
```

so that we have better control over the imports. This way, our profile would be:

```python
from nsls2tools.csx1.startup import *
```

Only!

I didn't add the tardis yet as it was all commented out.

ambarb:
Nice idea to import it.

Are you looking at 98-tardis-mu.py?  This is the one we use and these things are in `sd.baseline` so they have to be defined.

stuwilkins:
Thanks, just added. I was confused by 98 and 99 ... anyway it is in now.

stuwilkins:
@danielballan @tacaswell @stuartcampbell 

Is this what you were thinking when we discussed this with @ambarb a while ago in the conference room?

stuwilkins:
@ambarb if you are in, lets try and get this done Tuesday AM and see if it works, if it does then this is a model for SIX......

ambarb:
Sounds good.  I was able to get fix some of our custom plans so that the header information is populated.  Just need to fix the plot scaling. I will need some help from DAMA with that because I don't know how to pass that variable to `bec`. @stuartcampbell  can someone help me with this tomorrow?

stuartcampbell:
Someone should be able to helpout... @danielballan @licode is one of you available ?

nsls2tools/csx1/ophyd/areadetector.py
```diff
@@ -0,0 +1,133 @@
...
+    # This does nothing, but it's the right place to add code to be run
+    # once at instantiation time.
+    def __init__(self, *arg, readout_time=0.04, **kwargs):
+        self.readout_time = readout_time
+        super().__init__(*arg, **kwargs)
+
+    def pause(self):
+        self.cam.acquire.put(0)
+        super().pause()
+
```
stuwilkins: @ambarb have a look at lines 90 to 100. I am thinking that instead of having these __macros__ which set the FCCD acquisition time, we use this method on the class to do it. Then you would just do:

```python
fccd.set_acquisition(0.1)
```
at the ipython prompt. Curious what @tacaswell and @danielballan think, can this be poked from BS?

danielballan:
I really like how this is looking. I'm wondering whether the subpackages `bluesky`, `ophyd`, and `analysis`, are the organization we want to go with long term. I can imagine some cases where things might not fit neatly into just one. Would it be better to move all those modules in the top level (`csx`) package? No strong feelings here -- just brainstorming.

I think it _does_ make sense to have a `startup` subpackage so that we can `from nsls2utils.csx1.startup import *`.

stuwilkins:
Thanks @danielballan 

I was thinking that perhaps the distinction should be _devices_ and _plans_ rather than the actual package names? What do you think.

I like the idea of separating the locations of these things

danielballan:
I almost made that exact suggestion myself. Let's go with that.

Get Outlook for iOS<https://aka.ms/o0ukef>



On Wed, Sep 27, 2017 at 6:02 PM -0400, "Stuart Wilkins" <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks @danielballan<https://github.com/danielballan>

I was thinking that perhaps the distinction should be devices and plans rather than the actual package names? What do you think.

I like the idea of separating the locations of these things

stuwilkins:
@stuartcampbell @danielballan 

This is now a fairly stable config for CSX. I think it is ready for review comment. Could someone take a look. I am keen for this not to be some monster PR as well!

danielballan:
Haha, too late! But sometimes monster PRs are necessary -- the goal now should be to merge this promptly. I'll take a look.

nsls2tools/csx1/plans/suspenders.py
```diff
@@ -0,0 +1,36 @@
+from bluesky.suspenders import (SuspendBoolHigh,
+                                SuspendBoolLow,
+                                SuspendFloor,
+                                SuspendCeil,
+                                SuspendInBand,
+                                SuspendOutBand)
+
+from ophyd import EpicsSignal
+
+### Enable this line for standard GU operation
```
danielballan: Intentionally disabled? Perhaps this is awaiting the improved suspenders I've been talking about adding to bluesky, which you allow suspenders to block certain operations (triggering the FCCD while the beam is down) but allow others (moving a motor while the beam is down).

nsls2tools/csx1/startup/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .startup import *
+from .accelerator import *
+from .optics import *
+from .endstation import *
+from .detectors import *
+#from ..bluesky.xpcs import *
```
danielballan: Intentionally commented out? If so please remark on why.

nsls2tools/csx1/startup/settings.py
```diff
@@ -0,0 +1,95 @@
...
+sd.baseline = [theta, delta, gamma,
+               sx, say, saz,
+               cryoangle, sy, sz,
+               epu1, epu2,
+               slt1, slt2, slt3,
+               m1a, m3a,
+               nanop, tardis,
+               stemp, pgm]
+
+sclr.names.read_attrs=['name1','name2','name3','name4','name5','name6']  # TODO  WHAT IS THIS??? - Dan Allan
```
danielballan: See my comment in the code. ;- )

nsls2tools/csx1/startup/settings.py
```diff
@@ -0,0 +1,95 @@
...
+    m1a.z,
+    m1a.y,
+    m1a.x,
+    m1a.pit,
+    m1a.yaw,
+    m1a.rol,
+    m3a.x,
+    m3a.pit,
+    m3a.bdr,
+    # muR,  # TODO turn this back on when safe
```
danielballan: Is this safe to turn back on? I see that `muR` and `muT` are defined in `tardis.py` now so my guess is that uncommenting this would be safe to do.

nsls2tools/csx1/devices/lakeshore.py
```diff
@@ -0,0 +1,195 @@
...
+    A = Cpt(DelayGeneratorChan, '-Chan:A}DO:Dly')
+    B = Cpt(DelayGeneratorChan, '-Chan:B}DO:Dly')
+    C = Cpt(DelayGeneratorChan, '-Chan:C}DO:Dly')
+    D = Cpt(DelayGeneratorChan, '-Chan:D}DO:Dly')
+    E = Cpt(DelayGeneratorChan, '-Chan:E}DO:Dly')
+    F = Cpt(DelayGeneratorChan, '-Chan:F}DO:Dly')
+    G = Cpt(DelayGeneratorChan, '-Chan:G}DO:Dly')
+    H = Cpt(DelayGeneratorChan, '-Chan:H}DO:Dly')
+
+# Current-Voltage meter, driven in current mode
```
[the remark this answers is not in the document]
ambarb: In principle, we should not need a beter ioc.  It was commented out when we stopped using it.  (last used a year ago July).  With all of the software changes, I am not sure that the current IOC will work with this current bluesky device.  It is something I will test in the next days when there is no beam.

I think it was commented out because this information was not wished to be recorded or to prevent people from trying to use.  @danielballan What do you suggest for handling these kings of items (not regularly used).  I imagine that once we get a sample holder for this capability that allows us to use the loadlock, that this device will become used more often, but still not for every (or likely most) experiments.

nsls2tools/csx1/plans/suspenders.py
ambarb: Someone commented out for using the beamline during the shutdown.  Then all of it was commented out to prevent people from thinking that the suspenders were enabled.  Perhaps this should set up and define the suspender, but not install it.  We could print to the screen how to install the suspender.  THen the user can put that line in the local startup file that includes user specific `md`.   @stuwilkins @cmazzoli @wenhu ?  
Then we can re-evaluate things once the improved suspender capability is launched.

nsls2tools/csx1/startup/__init__.py
ambarb: We need to test this plan as it was something we did before the gs.DETS were removed. I believe it should be okay, but I think we might want to add some metadata to determine the state of the beam block and gain state.

nsls2tools/csx1/startup/settings.py
ambarb: It should be.  Right now the motors are not plugged in so no one can move it until I am ready to test it.  Which would mean that we would need to enable it.

@stuwilkins if I were to want to update this, should I:
* work on your fork once back at NSLS-II
* take my fork of nsls-ii-tool and fetch your fork onto mine?
* make a fork of your fork (which seems a bit complicated) but maybe this is a good way to force a LBS review of the request before the big merge back into the facility

nsls2tools/csx1/startup/settings.py
ambarb: @danielballan This was not my question.  But I do not see your answer in the code directly above or below.

ambarb: you mean 
```
+# This was imported in 00-startup.py #  used to generate the list: [thing.name for thing in get_all_positioners()]```

File: nsls2tools/csx1/plans/nanop.py [hunk 7]
```diff
@@ -0,0 +1,114 @@
...
+        x center
+    y_start : float
+        y center
+    npts : integer
+        number of points
+    probe_size : float
+        radius of probe in units of motors
+    overlap : float
+        fraction of probe overlap
+    ----------------------------------------------------------------
```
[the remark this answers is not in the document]
ambarb: Ok.  we can deleted the dashes and just make a note for the parameter.

Also, since these plans are technically CSX standard scan plans (and those that will go in this subdirectory in the future), how do we feel about adding the olog statement here so that the user does not need to decorate their overall plan? Everyone can comment: @stuwilkins @cmazzoli @wenhu @danielballan @tacaswell  @stuartcampbell  and other interested parties.

ambarb:
Add a logic test to the fccd configuration attrs to determine if the actuator for dark images is in or out.

danielballan:
This content will be moved into csxtools and/or back to the IPython profile.

@stuwilkins Do you think a *tools repo for your _program_ would be useful, or is it better to put things either in csxtools, sixtools, or scikit-beam?

tacaswell:
This has moved back to the CSX program for now.

PR #15, opened 2018-01-09
Added templating to configure_olog

jrmlhermitte:
Some beamlines use templating for olog, example:
https://github.com/NSLS-II-CMS/profile_collection/blob/master/startup/45-olog.py#L90

@danielballan

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -173,7 +173,9 @@ def configure_base(user_ns, broker_name, *,
     return list(ns)
 
 
-def configure_olog(user_ns, *, callback=None, subscribe=True):
+def configure_olog(user_ns, *, callback=None, subscribe=True,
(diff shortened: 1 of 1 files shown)


danielballan:
I provided a took for CMS and CHX to [pass in their `logbook_cb_factory` directly](https://github.com/NSLS-II/nslsii/blob/master/nslsii/__init__.py#L192), which I slightly prefer because it minimizes the interface between these two components and allows for other possible customizations, all wrapped into `logbook_cb_factory`.

jrmlhermitte:
so `partial(logbook_cb_factory(**kwargs)` correct?

danielballan:
Yes, that should be passed to `configure_olog`'s `callback` argument.

danielballan:
So many layers in Olog!! hahah

jrmlhermitte:
Yeah partial not necessary. I overthought this one lol

PR #18, opened 2018-01-24
Add xspress3 from hxntools

jrmlhermitte:
(half the time making this was properly spelling "grep -ri xspress3", why the "s"? hehe)
This is a beginning, taking from hxntools. There are no tests, so probably a good idea would be to test this at CHX. I'll contact @yugangzhang.

Do not merge until tested.

### DIFF
--- a/nslsii/detectors/trigger_mixins.py
+++ b/nslsii/detectors/trigger_mixins.py
@@ -1,0 +1,184 @@
+import time as ttime
+import itertools
+import logging
+
+from ophyd.device import (DeviceStatus, BlueskyInterface, Staged,
+                          Component as Cpt, Device)
+from ophyd import (Signal, )
+from ophyd.areadetector.filestore_mixins import FileStoreIterativeWrite
+
+
+from .utils import ordered_dict_move_to_beginning
+
+logger = logging.getLogger(__name__)
+
+
+class TriggerBase(BlueskyInterface):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # If acquiring, stop.
+        self.stage_sigs[self.cam.acquire] = 0
+        self.stage_sigs[self.cam.image_mode] = 'Multiple'
+        self._acquisition_signal = self.cam.acquire
+
+        self._status = None
+
+
+class HxnModalSettings(Device):
+    mode = Cpt(Signal, value='internal',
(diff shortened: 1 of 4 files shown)


danielballan:
Garth mentioned today that he customized this code at SRX via subclassing. We should incorporate that as well, I think. Not necessarily in the first pass / same PR. If this PR works at HXN and CHX, I would say that's a good enough scope to get started.

jrmlhermitte:
tested at chx. had to add some files. it's good to go. 
Just one more question. In this file:
https://github.com/NSLS-II/nslsii/pull/18/files#diff-c7ed285800e7e446373238a9653a9645

there are three convenience routines: `makedirs`, `ordered_dict_move_to_beginning` and `make_filename_add_subdirectory`. I'm not sure where to put them. Should we move them one level up or keep them `detectors` specific utils?
@danielballan

jrmlhermitte:
We should put a refernence to where this code came from in the commit just in case, and acknowledge @licode  and @tacaswell .

jrmlhermitte:
renamed `HxnModalBase` `HxnModalSettings` `HxnModalTrigger` to  `ModalBase` `ModalSettings` `ModalTrigger`

danielballan:
Great! This has been a long time coming. Thanks to @yugangzhang for taking the time to help us do this in a sustainable way that will help the whole facility.

>  I'm not sure where to put them.

I think their current location is fine. We can always move and shim if we change our minds.

> We should put a refernence to where this code came from in the commit just in case, and acknowledge @licode and @tacaswell.

The most important author here may be klauer, but @tacaswell can correct me if I'm wrong. Credit can be given by amending these commits with `git commit --amend --author=klauer@bnl.gov`, by putting a comment at the top of the relevant file(s), or (to be really hard-core about it) by pulling in the git history of the hxntools repo and filtering out the parts that don't affect these files. I think the ship has sailed on the purist approach -- not worth the trouble now that we have this working.

jrmlhermitte:
sounds good. i added them in the log for the latest commit; i kept my author because i should be to blame if this didnt work :-P.

looks good to me, are you ok with this @danielballan ?

tacaswell:
I didn't see this go by last week :/

I am not sure that any of the 'modal' classes should have made it out of the hxntools as they are to support the very specific way that HXN is configured.  Any classes we don't need to make the exspress3 work at ex SRX should be deleted.

PR #22, opened 2018-05-10
updated configure_base for integrating ArchiverEventSources

malitsky:
updated configure_base for integrating ArchiverEventSources

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -179,6 +179,11 @@ def configure_base(user_ns, broker_name, *,
     import bluesky.simulators
     import_star(bluesky.simulators, ns)
 
+    from databroker.eventsource.archiver import *
+    for name in ArchiverReader.archiver_names():
+        aes = ArchiverReader.named(name)
+        db.add_event_source(aes)
+    
     user_ns.update(ns)
     return list(ns)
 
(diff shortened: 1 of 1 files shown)


nslsii/__init__.py
```diff
@@ -179,6 +179,11 @@ def configure_base(user_ns, broker_name, *,
     import bluesky.simulators
     import_star(bluesky.simulators, ns)
 
+    from databroker.eventsource.archiver import *
```
tacaswell: Please do not use `import *`

tacaswell:
I think we want to merge the base configuration into `db.named` instead of having a separate lookup system.

Do we want to allow `configure_base` to make additional PVs to be included?  That seems like something that will be in flux for a while ?   For now I guess just walking the `db.EventSources` and checking the type is the easiest way to get the right EventSource?  I suspect we want to also include some notion of 'name' in the config when creating the EVentSource objects and have some access to that from the DataBroker, ex `db.ES['archiver']` or `db.ES['raw']` ?

danielballan:
:+1: for putting the list of PVs into the databroker configuration file.

malitsky:
integration is implemented with the Broker API using the databroker configuration file and db.ES  as demonstrated by the following script and configuration file:

https://github.com/malitsky/nsls2/blob/master/areventsource/notebooks/archiver-reader.db.5.13.18.ipynb
https://github.com/malitsky/nsls2/blob/master/areventsource/notebooks/csx_with_archivers.yml

PR #36, opened 2018-09-10
ENH: add V33 compatible version of Stats plugin

tacaswell:
PV name changes due to https://github.com/areaDetector/ADCore/pull/333

Work of @mrakitin, @malitsky, and Oksana Ivashkevych.


mrakitin:
Good catch of the source of the changes. Here is some explanation of the changes: https://github.com/areaDetector/ADCore/pull/333/files#diff-781a53e648f3df8d16a08ec083b04bf4R121.

PR #40, opened 2018-09-28
ENH: add EpicsSignal/RO precision classes

mrakitin:
Originally implemented at [SMI](https://github.com/NSLS-II-SMI/profile_collection/blob/70909a1916168872d98795cdd219ab18c01d5c39/startup/09-machine.py#L30-L47) by @danielballan. Recently adopted at [XPD](https://github.com/NSLS-II-XPD/profile_collection/blob/cb9e79e9a19fcefb73935020d1a27590a68d1d78/startup/12-rga.py#L5-L22).

### DIFF
--- a/nslsii/devices.py
+++ b/nslsii/devices.py
@@ -3,10 +3,30 @@ import datetime
 from ophyd import (Device, Component as Cpt,
                    EpicsSignal, EpicsSignalRO, DeviceStatus)
 
-_time_fmtstr = '%Y-%m-%d %H:%M:%S'
+
+class EpicsSignalOverridePrecRO(EpicsSignalRO):
+    def __init__(self, *args, precision=4, **kwargs):
+        self._precision = precision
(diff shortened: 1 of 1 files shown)


danielballan:
If I recall correctly, I did this as a dirty hack instead of just setting the IOC's precision. This should probably fixed the "right way" and not merged, unless there is some reason that it can't be fixed in the IOC.

mrakitin:
Agree on that. I'll raise the question about fixing it at the IOC level next time I visit XPD.

mrakitin:
xref: https://3.basecamp.com/3733838/buckets/4066459/todos/1355796496

mrakitin:
This is fixed on the IOC level at XPD. Will test it next time I visit.

mrakitin:
Closing without merging. Corresponding changes to the XPD profile are here: https://github.com/NSLS-II-XPD/profile_collection/pull/3.

PR #41, opened 2018-10-02
Add a Eurotherm device

awalter-bnl:
This PR will add a new Eurotherm device to replace the seperate Eurotherm devices in the [PDF](https://github.com/NSLS-II-PDF/profile_collection/blob/9753a08660d83071fadeccf82686c144599ef91f/startup/11-temperature-controller.py#L41) and [XPD](https://github.com/NSLS-II-XPD/profile_collection/blob/cb9e79e9a19fcefb73935020d1a27590a68d1d78/startup/11-temperature-controller.py#L41) profiles.

## Description
This will add a replacement to Eurotherm device for the XPD and PDF beamlines (and any others who will use these devices in the future). It resolves an issue where the IOC indicates to Ophyd that it is 'done moving' when it first reaches the setpoint temp not after it has reached eqilibrium at the setpoint temp. This is doen using the existing `tolerance` property of the EpicsSignalPositioner, of which the device is a child, and by adding to new properties, `equilibrium_time` and `timeout`. The latter is passed into the status object to indicate how long to wait for equilibrium before raising a `TimeOutError`. `equilibrium_time` is used to indicate how long the temperature should remain with `tolerance` of the setpoint before it is considered at equilibrium. 

## Motivation and Context
The Eurotherm device has the tendency to overshoot the temperature setpoint prior to ataining equilibrium. The IOC indicates that it is done changing temperature when the setpoint is first reached, this means that bluesky is moving onto the next point in a scan when the temperature has not reached equilibrium. The controls group have indicated that this can not be resolved at the IOC, or lower, level therefore I am implementing this solution in the ophyd layer.

## How Has This Been Tested?
This has yet to be tested, once it has been reviewed and approved I will test at either XPD or PDF on the device prior to releasing the Do Not Merge label and allowing for merging.


awalter-bnl:
I am still not 100% clear on how status objects work. In this case I am not 100% sure that:
1. If the `if` statement on line 75 fails will it return to checking the `if` statement on line 73 or not.
2. Will the `timeout` be implemented effectively (i.e. when does the timing of the timeout start).

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,85 @@
...
+    # This approach mirrors that used for tolerance in ophyd/signal/Signal
+    @property
+    def equilibrium_time(self):
+        '''The time that a temperature should be within `tolerance` for it to
+        be considered at equilibrium.
+        '''
+        return self._equilibrium_time
+
+    @equilibrium_time.setter
+    def equilibrium_time(self, equilibrium_time):
```
mrakitin: We have 3 different objects called `equilibrium_time` - a kwarg in the `__init__(...)`, a function, and a kwarg for this function. Will we see any name collisions?

awalter-bnl: That is an excellent question @mrakitin. this structure (and the 3 different objects with the same name) mirrors what is found for `tolerance` in [ophyd/signal/Signal](https://github.com/NSLS-II/ophyd/blob/112c88968e1965803868195818d4ae122b742ee0/ophyd/signal.py#L91) and does not create collisions there. 

I like using the same name as it makes things a little easier to follow (IMO) but am also open to changing it it is considered to risky.

mrakitin: OK, makes sense to me, thanks for the reference, @awalter-bnl!

tacaswell: We already have a 'settle_time' on positioners, we probably should re-use that name here.

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,85 @@
...
+    def set(self, value):
+        # define some required values
+        set_value = value
+        status = DeviceStatus(self, timeout=self.timeout)
+
+        # set up the done moving indicator logic
+        def status_indicator(value):
+            if abs(value - set_value) < self.tolerance:
+                sleep(self.equilibrium_time)
+                if abs(value - set_value) < self.tolerance:
```
tacaswell: This will test the same value twice.

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,85 @@
...
+                    status._finished()
+
+        # Start the move.
+        self.put(set_value)
+
+        # subscribe to the read value to indicate the set is done.
+        self.position.subscribe(status_indicator)
```
tacaswell: need to make sure you also un-subscribe this.

nslsii/temperature_controllers.py
awalter-bnl: This is different from a settle time. A settle time is how long to wait after motion is complete to wait before moving on. This is how long to wait before rechecking that the temperature is within the tolerance range, and hence the temperature is 'stable'. I htink it makes sense to use a different term here as they are different values.

nslsii/temperature_controllers.py
awalter-bnl: It is suppose to test the same value twice, after waiting `equilibrium_time` between tests. This is suppose to check if the temperature is stable, and within the tolerance, over equilibrium_time.

nslsii/temperature_controllers.py
awalter-bnl: I will look into this, not sure where the unsubscribe should occur but I am sure I can work it out.

tacaswell:
https://github.com/NSLS-II/caproto/pull/343 <- that may be useful for testing this (modulo fixing some names).

nslsii/temperature_controllers.py
tacaswell: I mean, `value` does not change here.

I am also concerned that this will block the libca / caproto event loop (the two button shutter also has this issue..).

awalter-bnl: I am confused, this matches the structure [here](https://github.com/NSLS-II-ESM/profile_collection/blob/3217114727d94a1343a53eda6e9876f0d621169b/startup/10-machine.py#L156) where I assume that subscribing this to the `position` attribute updates `value`. 

Is the mistake I am making that I need to replace the `value` in line 75 with `self.position()`?

Can anyone suggest a better solution to resolve the blocking of libca / caproto event loop?
If not I will do some research.

awalter-bnl:
thanks for the link @tacaswell I will check it out.

awalter-bnl:
I have made the changes requested by @tacaswell except:
1. changing `equilibrium_time` to `settle_time`. @tacaswell are you convinced that they are different enough to warrant different names?
2. the issue of blocking of libca/caproto event loops, I am still investigating the issue in order to find possible solutions (comments/suggestions on this are welcomed)

awalter-bnl:
The latest commit is after I have tested against the simulated temp controller @tacaswell wrote in a caproto PR [here](https://github.com/NSLS-II/caproto/pull/343/files). My next step is to add a test using the caproto simulated temp controller.

nslsii/temperature_controllers.py
```diff
@@ -13,74 +12,54 @@ class Eurotherm(EpicsSignalPositioner):
...
+            if abs(value - set_value) < self.tolerance.get():
+                if initial_timestamp:
+                    if ((timestamp - initial_timestamp) >
+                                          self.equilibrium_time.get()):
+                        status._finished()
+                        self.readback.clear_sub(status_indicator)
+                else:
+                    initial_timestamp = timestamp
+            else:
+                inband = False
```
tacaswell: rouge `inband` left.

nslsii/temperature_controllers.py
```diff
@@ -13,74 +12,54 @@ class Eurotherm(EpicsSignalPositioner):
...
         # set up the done moving indicator logic
-        def status_indicator(value):
-            if abs(value - set_value) < self.tolerance:
-                sleep(self.equilibrium_time)
-                if abs(self.position - set_value) < self.tolerance:
-                    status._finished()
-                    self.position.clear_sub(status_indicator)
+        def status_indicator(value, timestamp, **kwargs):
+            nonlocal initial_timestamp
+            if abs(value - set_value) < self.tolerance.get():
```
tacaswell: might be worth grabbing the tolerance and equilibrium time values outside of this function.  It will be marginally faster and changing these things while it is settling is a bit weird to support.

tacaswell:
Might be worth adding some locking so that if there is already a set in progress, we can not start a second one?  Have to be very careful about it getting cleaned up in stop though. 

We should also add a stop method that sets it to it's current value.  This isn't _great_ but seems like the best way to stop a temperature controller.

awalter-bnl:
@tacaswell I have made the requested changes now to work on the test function !-)

awalter-bnl:
@tacaswell thanks for the link it is very helpful, I have a question about:
> I think it may be worth moving the example IOC into this repo to prevent weird inter-project dependencies.

What do you mean by 'example IOC'? Do you mean just a copy of the `pytest_fixture` found in your link above, the `pytest_fixture` and  a copy of `run_example_ioc()` or the entire simulated thermo IOC from [caproto PR # 343](https://github.com/NSLS-II/caproto/pull/343). 

It is not 100% clear to me which you mean.

tacaswell:
> What do you mean by 'example IOC'? Do you mean just a copy of the `pytest_fixture` found in your link above, the `pytest_fixture` and a copy of `run_example_ioc()` or the entire simulated thermo IOC from [caproto PR # 343](https://github.com/NSLS-II/caproto/pull/343).

I importing the pytest fixtures and helpers in probably ok (those are likely to not change in breaking ways anymore due to their heavy use in caproto testing already), but we should copy the IOC from https://github.com/NSLS-II/caproto/pull/343 over in entirety (as it is just an example and nothing makes sure we don't change some small detail of the example to make it a better example that would make it badly suited for the tests here).

awalter-bnl:
I have removed the DO NOT MERGE tag, as it was discussed this morning that an update to this is imminent and wanted this included. I will add a test function via a furture PR (the current code has been tested manually using the new simulated temp controller).

awalter-bnl:
I have added a test for the new device, this should be reviewed before merging.

awalter-bnl:
closed and re-opened to start a Travis CI build

awalter-bnl:
FYI I am having some issues getting the test to run in Travis. It runs locally without issue, I am trying to work through the problem now.

awalter-bnl:
OK so I have now managed to get the test to pass, So this is ready for review. (FYI the issue was that I didn't have caproto as a requirement.)

awalter-bnl:
thanks @mrakitin for letting me know about the requirements list ordering. That is now fixed

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,81 @@
+from ophyd import DeviceStatus, Device, Component as Cpt, EpicsSignal, Signal
+
+
+class Eurotherm(Device):
+    '''This class is used for integrating with Eurotherm controllers.
+
+    Parameters
+    ----------
+    PV_prefix : str.
```
mrakitin: Why not to use a lower-case name here, i.e. `pv_prefix`?

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,81 @@
...
+        The PV prefix that is common to the readback and setpoint PV's.
+    tolerance : float, optional.
+        The range of temperature within which it can be considered constant.
+        Default is 1.
+    EquilibriumTime : float, optional.
```
mrakitin: I think it should match the Component name below: `equilibrium_time`.

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,81 @@
...
+        The time (in seconds) that a temperature should be within `tolerance`
+        for it to be considered at equilibrium. Default is 5 seconds.
+    Timeout : float, optional.
```
mrakitin: This also should be lower-case: `timeout`.

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,81 @@
...
+    # Setup some new signals required for the moving indicator logic
+    equilibrium_time = Cpt(Signal, value=5)
+    timeout = Cpt(Signal, value=500)
+    tolerance = Cpt(Signal, value=1)
+
+    # Add the readback and setpoint components
+    setpoint = Cpt(EpicsSignal, 'SP')
+    readback = Cpt(EpicsSignal, 'I')
+
+    _set_lock = False
```
mrakitin: I wonder is this `_set_lock` should go under `__init__`?

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,81 @@
...
+    EquilibriumTime : float, optional.
+        The time (in seconds) that a temperature should be within `tolerance`
+        for it to be considered at equilibrium. Default is 5 seconds.
+    Timeout : float, optional.
+        The maximum time (in seconds) to wait for the temperature to reach
+        equilibrium before raising a TimeOutError. Default is 30 seconds.
+    '''
+
+    def __init__(self, PV_prefix, **kwargs):
+        super().__init__(PV_prefix, **kwargs)
```
mrakitin: In beamline profiles this block normally goes after all components are defined. Maybe worth to move it below.

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,81 @@
...
+        # grab these values here to avoidmutliple calls.
+        equilibrium_time = self.equilibrium_time.get()
+        tolerance = self.tolerance.get()
+
+        # set up the done moving indicator logic
+        def status_indicator(value, timestamp, **kwargs):
+            nonlocal initial_timestamp
+            if abs(value - set_value) < tolerance:
+                if initial_timestamp:
+                    if ((timestamp - initial_timestamp) > equilibrium_time):
```
mrakitin: You may omit the outer parentheses here.

nslsii/tests/temperature_controllers_test.py
```diff
@@ -0,0 +1,50 @@
...
+    '''
+
+    stdout = subprocess.PIPE
+    stdin = None
+
+    # Start up an IOC based on the thermo_sim device in caproto.ioc_examples
+    ioc_process = subprocess.Popen([sys.executable, '-m',
+                                    'caproto.tests.example_runner',
+                                    'caproto.ioc_examples.thermo_sim'] +
+                                   list([]), stdout=stdout, stdin=stdin,
```
mrakitin: Why do you need to sum an empty list here?

nslsii/temperature_controllers.py
awalter-bnl: good point, I have made this change.

nslsii/temperature_controllers.py
awalter-bnl: I am not sure this is true, at least [here](https://github.com/NSLS-II-SIX/profile_collection/blob/77ee29dcd530aa95c61401efdc94ad7c6c21bba8/startup/21-areadetector.py#L22) it is not. @danielballan  or @tacaswell  do we have a standard regarding if the `__init__` comes before or after `component` definitions? If so I will change this to match the standard.

nslsii/temperature_controllers.py
awalter-bnl: done

nslsii/tests/temperature_controllers_test.py
awalter-bnl: you are correct this is not needed and removed.

awalter-bnl:
@mrakitin I have made the requested changes, but there is one outstanding comment to be resolved above.

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,80 @@
...
+
+        initial_timestamp = None
+
+        # grab these values here to avoidmutliple calls.
+        equilibrium_time = self.equilibrium_time.get()
+        tolerance = self.tolerance.get()
+
+        # set up the done moving indicator logic
+        def status_indicator(value, timestamp, **kwargs):
+            nonlocal initial_timestamp
```
danielballan: If initial_timestamp is defined here (initialized to None) instead of above, would that work? If so, that’s much simpler than invoking nonlocal.

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,80 @@
...
+            if abs(value - set_value) < tolerance:
+                if initial_timestamp:
+                    if (timestamp - initial_timestamp) > equilibrium_time:
```
danielballan: This assumes that we will happen to get a new reading soon enough that we aren’t left waiting to long after the equilibrium time has passed. It could hang forever. Should instead start a timer that will finish the status object unless interrupted by seeing an incompatible reading.

nslsii/temperature_controllers.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Parameters
+    ----------
+    pv_prefix : str.
+        The PV prefix that is common to the readback and setpoint PV's.
+    '''
+
+    def __init__(self, pv_prefix, **kwargs):
+        super().__init__(pv_prefix, **kwargs)
+        self._set_lock = False
```
danielballan: We should maybe we using real locks for this but this is a common enough pattern in ophyd so I think we can circle back for that later.

awalter-bnl: I am happy to circle back later.

awalter-bnl:
I do not see a show-stopping bug here, @danielballan can you elaborate?

nslsii/temperature_controllers.py
danielballan: Let me make sure I understand correctly what this supposed to do:
1. Set the setpoint and subscribe to updates from the readback.
3. As soon as we see a reading within the tolerance of the setpoint, start a timer (effectively).
4. If we see a reading outside the tolerance, restart the timer.
5. If we are another reading inside the tolerance and if the equilibration time has passed, we are done.

The bug I see is that that last reading could be a long time in coming, and we might be well beyond the equilibration time by then. At best we waste little bits of experiment time, at worst something times out and we get a false failure. I suggest reworking 5 so that we are done immediately when the equilibration time expires, not waiting for the first reading after that time.

danielballan: Let me make sure I understand correctly what this supposed to do:
1. Set the setpoint and subscribe to updates from the readback.
3. As soon as we see a reading within the tolerance of the setpoint, start a timer (effectively).
4. If we see a reading outside the tolerance, restart the timer.
5. If we are another reading inside the tolerance and if the equilibration time has passed, we are done.

The bug I see is that that last reading could be a long time in coming, and we might be well beyond the equilibration time by then. At best we waste little bits of experiment time, at worst something times out and we get a false failure. I suggest reworking 5 so that we are done immediately when the equilibration time expires, not waiting for the first reading after that time.

danielballan: Let me make sure I understand correctly what this supposed to do:
1. Set the setpoint and subscribe to updates from the readback.
3. As soon as we see a reading within the tolerance of the setpoint, start a timer (effectively).
4. If we see a reading outside the tolerance, restart the timer.
5. If we are another reading inside the tolerance and if the equilibration time has passed, we are done.

The bug I see is that that last reading could be a long time in coming, and we might be well beyond the equilibration time by then. At best we waste little bits of experiment time, at worst something times out and we get a false failure. I suggest reworking 5 so that we are done immediately when the equilibration time expires, not waiting for the first reading after that time.

awalter-bnl: So that was the original structure that I had in the PR, and after comments from @tacaswell above, and 1-1 discussions with him I changed it to this structure.
To elaborate:
What I initially had was a `sleep(equilibrium_time)` after the first time we were in-band and then I checked after that time and exited if we where still in-band. This did not work as the second `self.readback.get()` returned the same value before and after the `sleep`. I discussed this issue with @tacaswell and he suggested the current implementation as the solution. 

If someone has a better solution than using `sleep` to allow for the wait I am happy to look at implementing it, but we have tried using `sleep` and that did not work.

nslsii/temperature_controllers.py
tacaswell: No, we need the closure so the next time the calback runs we can check the value.

nslsii/temperature_controllers.py
tacaswell: The eurothrem updates on a clock (which is a PV we can control).

You can not sleep in this callback because it blocks libca processing any other updates.

danielballan: That was the key fact missing from this discussion. In that case we waste at most one click cycle. Downgrading from “show-stopping” to “not optimal.”

I was suggesting sleeping in anew thread,perhaps via a Timer object, exactly as we do for settle_time. 👎on inventing a new name for what is the temperature controller equivalent of settle time because that breaks hardware generality a little but it’s not a hill I’ll die on.

danielballan: That was the key fact missing from this discussion. In that case we waste at most one click cycle. Downgrading from “show-stopping” to “not optimal.”

I was suggesting sleeping in anew thread,perhaps via a Timer object, exactly as we do for settle_time. 👎on inventing a new name for what is the temperature controller equivalent of settle time because that breaks hardware generality a little but it’s not a hill I’ll die on.

danielballan: That was the key fact missing from this discussion. In that case we waste at most one click cycle. Downgrading from “show-stopping” to “not optimal.”

I was suggesting sleeping in anew thread,perhaps via a Timer object, exactly as we do for settle_time. 👎on inventing a new name for what is the temperature controller equivalent of settle time because that breaks hardware generality a little but it’s not a hill I’ll die on.

awalter-bnl: Thanks for the feedback @danielballan, hopefully with time I will get better at specifying the issues in a way that is clearer to you.

In regard to the name `equilibrium_time` this was also raised by  @tacaswell .

> **tacaswell 16 days ago Member**
> We already have a 'settle_time' on positioners, we probably should re-use that name here.

>**awalter-bnl 16 days ago Member**
>This is different from a settle time. A settle time is how long to wait, after motion is complete, before >moving on. This is how long to wait before rechecking that the temperature is within the tolerance >range, and hence the temperature is 'stable'. I think it makes sense to use a different term here as >they are different values.

So far I haven't heard back from @tacaswell regarding if he thinks my argument here makes sense, I prefer a different name because I am not convinced it is the temperature controller equivalent of `settle_time`, but will change it if this argument does not convince others.

tacaswell:
The timer class @danielballan is talking about is https://docs.python.org/3.6/library/threading.html#timer-objects

The logic would be to if it is in-band and there is no timer arm it if there is a timer move on, if we ever go out of band cancel it and reset.

The callback on the timer would flip the status object and pulls the callback off the readback.

I am 70/30 on which way is better.  Dan is right in principle we may never see the updated value, but in practice this IOC runs with the SCAN value set to some period.

tacaswell:
To be clear, the 70 is towards Dan's suggestion.

awalter-bnl:
Ok so just to be 100% clear, this would also only work if I remove the `timeout` kwarg [here](https://github.com/NSLS-II/nslsii/blob/0362505164865767ec34ae1cc9e2bf6c40f7783b/nslsii/temperature_controllers.py#L45).

As this is essentially placing the timeout logic into `status_indicator`? am I correct.

The only issue I sort of see with this is that it 'restarts' the `timer` everytime it first goes in-band then we are constantly moving the start time for the `timeout`. So in principle it could take much longer than `timeout` before raising an error. This is a subtle change that won't be immediately apparent, and could cause some confusion.

tacaswell:
This is orthogonal to the timeout on the status object its self (which is set up at https://github.com/NSLS-II/ophyd/blob/b907a8c967a8eb4353666b8dd632d87bd1716e2c/ophyd/status.py#L66-L70 ).  That logic is "if after the time out time we have _not_ otherwise completed, unconditionally clean up and fail".  This as  separate time to succeed.

nslsii/temperature_controllers.py
awalter-bnl: @danielballan just so I can circle back to this later, when you specify 'real lock' do you mean a threading.Lock instacne as described [here](https://docs.python.org/3.6/library/threading.html#timer-objects)?

PR #42, opened 2018-10-04
ENH: add QuadEMV33 class with StatsPluginV33

mrakitin:
This is used at XPD with the AD33 StatsPlugin: https://github.com/NSLS-II-XPD/profile_collection/pull/1.

### DIFF
--- a/nslsii/ad33.py
+++ b/nslsii/ad33.py
@@ -14,6 +14,7 @@ from ophyd.areadetector.plugins import PluginBase
 from ophyd.areadetector.trigger_mixins import TriggerBase, ADTriggerStatus
 from ophyd.device import DynamicDeviceComponent as DDC, Staged
 from ophyd.signal import (Signal, EpicsSignalRO, EpicsSignal)
+from ophyd.quadem import QuadEM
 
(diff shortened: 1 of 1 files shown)


mrakitin:
I realized it may not be necessary with the profile-level fix: https://github.com/NSLS-II-XPD/profile_collection/pull/1/commits/983f7836e6a183c1ab4479b36fa1eb9c3915dc82.

tacaswell:
sub-classing may be a better solution is both cases.

mrakitin:
Good idea, @tacaswell! Do we still want it here in nslsii?

mrakitin:
Finally, added QuadEMPort to avoid having it in the BL profiles.

mrakitin:
I tested this fix on xf28id2-ws6 with the source-installed code from this PR branch into collection-2018-3.2, works great after the recent fixes with https://github.com/NSLS-II-XPD/profile_collection/pull/1/files. I think we should merge it soon and tag a new version of nslsii, build it, generate a minor env files (2018-3.3), and push to all XPD's machines, since they requested to do it for the experiments.

PR #43, opened 2018-10-17
Adding .yml file and enc file for Travis integration and doc creation

awalter-bnl:
This PR is designed to add a Travis.yml file and a github_deploy_key file to this repo in order to setup the Travis CI.


.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+cache:
+  directories:
+    - $HOME/.cache/pip
+    - $HOME/.ccache  # https://github.com/travis-ci/travis-ci/issues/5853
+
+install:
+  # Install this package and the packages listed in requirements.txt.
+  - pip install .
+  # Install extra requirements for running tests and building docs.
+  - pip install -r requirements-dev.txt
```
mrakitin:
```suggestion
```

danielballan: I love this new GH feature. I think you should probably also remove the line above with the comment.

.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+
+env:
+   global:
+    # Doctr deploy key for NSLS-II-SIX/sixtools
```
danielballan: The mention of SIX seems like an accidental copy/paste stowaway.

.travis.yml
awalter-bnl: I haven't seen that feature before, but it does look good. I will have to take a look.

thanks @mrakitin for indicating the issue, I have removed this in the latest commit.

.travis.yml
awalter-bnl: You are correct @danielballan this has been resolved in the latest commit

awalter-bnl:
Thanks @danielballan and @mrakitin for the quick review of this, much appreciated.

danielballan:
Wait, it seems like the right solution is not to remove the line installing `requirements-dev.txt` but to add a `requirements-dev.txt` file to install. The build is failing because several dev-related packages are not installed.

.travis.yml
```diff
@@ -0,0 +1,26 @@
+language: python
+python:
+  - 3.6
+cache:
+  directories:
+    - $HOME/.cache/pip
+    - $HOME/.ccache  # https://github.com/travis-ci/travis-ci/issues/5853
+
+install:
```
mrakitin: I think the packages for Travis should be added as well: `coverage`, `codecov`, `flake8`.
```suggestion
install:
  - pip install coverage codecov flake8
```

awalter-bnl:
I was getting there @mrakitin and @danielballan  but thanks for the heads up. I independently started @danielballan's solution.

danielballan:
Why 80? 79 is the standard.

awalter-bnl:
@danielballan I was just seeing if that was the issue, it turns out that wasn't the issue, it was flake8 in general

danielballan:
Ah, gotcha.

awalter-bnl:
So I asked the question on slack but will also add it here. The current failure is that the existing code in the repo does not meet flake8 (i.e. pep8) standards and is also failing coverage. Should I go through and fix all of these issues individually? (there is quite a few, I think most files in the repo are failing).

danielballan:
I vote for ignoring coverage but fixing flake8. It's a young project, might as well start off on the right foot.

As a general comment, I am against bombing old projects with style changes because it obscures the `git credit` history and risks introducing bugs. So, for example, pandas and IPython occasionally get hit with PRs that suggest ~1000s of lines of style "fixes" and I think they are right to decline them. But I think _new_ projects should start with flake8-compliance and young-ish projects like bluesky that _mostly_ comply can be fixed to literally comply with little downside.

nslsii/common/ipynb/info.py
```diff
@@ -3,10 +3,9 @@
 import platform
 from IPython.display import HTML
 
+
 def get_sys_info():
     """Display info on system and output as nice HTML"""
-    spacer = "<tr><td>&nbsp;</td><td>&nbsp;</td></tr>"
-
```
mrakitin: Why was it deleted?

awalter-bnl: It was unused according to flake8

awalter-bnl:
Appears now that everything is passing flake8, it is just the coverage that is failing.

nslsii/common/ipynb/info.py
mrakitin: Oh, got it. Thanks.

awalter-bnl: > F841 local variable 'spacer' is assigned to but never used

was the error

danielballan:
Because there are no tests to run and no docs to build, those commands are failing in Travis. I think we should leave all the requirements in place for our future use but comment the lines that attempt to run the tests and build the docs until we have tests to run and docs to builds. Sensible?

awalter-bnl:
PR #41 has a test in it which is why I added CI to this repo in the first place.

I am not against commenting out the build docs line however

PR #44, opened 2018-10-18
Add ipynb test

mrakitin:
I intended to add this test to make #43 pass the Travis test checks, but it was merged earlier than I pushed. So, staging it here.


awalter-bnl:
nice smoke test @mrakitin

mrakitin:
Thanks for reviewing and merging, @awalter-bnl!

Issue #45, opened 2018-10-19
Update the Eurotherm device with a Timer and real lock


awalter-bnl:
There are 2 tasks required to be completed on nslsii.temperature_controllers.Eurotherm

1. It has been requested to add a `threading.Timer` instance that runs inside `status_indicator` to ensure that the `status_indicator` callback is removed from self.readback after a timeout to prevent hanging. 

2. it has been requested to replace `self._set_lock` with a threading.Lock instance (I think this is what is meant by 'real lock').

PR #46, opened 2018-10-26
Update eurotherm

awalter-bnl:
## Description of issue
This PR is a response to issue #45. Essentially we can get into the situation where a callback 'hangs' because after a timeout error it is not removed from the readback signal. A second non-critical issue was raised, which is that we should be using threading.Lock objects instead of boolean objects as locks.
## Resolution
1. It adds a `threading.Timer` object that is used to do the timeout loop (instead of using the timeout in DeviceStatus).
2. it replaces the previous 'boolean' lock with a `threading.lock`.

The purpose of (1) is that currently the callback is not removed from readback when the timeout error is raised. So we could get into the situation whereby we never remove the callback. Moving the timeout process to a `threading.Timer` object associated allows the callback to be removed from readback after the timeout.
## Testing
The built in Travis pytest was updated to test these additions, including ensuring that the callback is removed on a timeout error.

### DIFF
--- a/nslsii/temperature_controllers.py
+++ b/nslsii/temperature_controllers.py
@@ -24,6 +24,11 @@ class Eurotherm(Device):
         super().__init__(pv_prefix, **kwargs)
         self._set_lock = threading.Lock()
 
+        # defining these here so that they can be used by `set` and `start`
+        self._status = None
+        self._cb_timer = None
+        self._cid = None
+
     # Setup some new signals required for the moving indicator logic
     equilibrium_time = Cpt(Signal, value=5)
     timeout = Cpt(Signal, value=500)
@@ -42,7 +47,7 @@ class Eurotherm(Device):
 
         # define some required values
         set_value = value
-        status = DeviceStatus(self)
+        self._status = DeviceStatus(self)
 
         initial_timestamp = None
 
@@ -52,28 +57,27 @@ class Eurotherm(Device):
 
         # setup a cleanup function for the timer, this matches including
         # timeout in `status` but also ensures that the callback us removed.
-        def timer_cleanup(status):
+        def timer_cleanup():
             print('Set of {} timed out after {} s'.format(self.name,
                                                           self.timeout.get()))
             self._set_lock.release()
             self.readback.clear_sub(status_indicator)
-            status._finished(success=False)
+            self._status._finished(success=False)
 
-        cb_timer = threading.Timer(self.timeout.get(), timer_cleanup,
-                                   args=(status,))
+        self._cb_timer = threading.Timer(self.timeout.get(), timer_cleanup)
 
         # set up the done moving indicator logic
         def status_indicator(value, timestamp, **kwargs):
             # add a Timer to ensure that timeout occurs.
-            if not cb_timer.is_alive():
-                cb_timer.start()
+            if not self._cb_timer.is_alive():
+                self._cb_timer.start()
 
             nonlocal initial_timestamp
             if abs(value - set_value) < tolerance:
                 if initial_timestamp:
                     if (timestamp - initial_timestamp) > equilibrium_time:
-                        status._finished()
-                        cb_timer.cancel()
+                        self._status._finished()
+                        self._cb_timer.cancel()
                         self._set_lock.release()
                         self.readback.clear_sub(status_indicator)
                 else:
@@ -88,11 +92,14 @@ class Eurotherm(Device):
         self._cid = self.readback.subscribe(status_indicator)
 
         # hand the status object back to the RE
-        return status
+        return self._status
 
     def stop(self):
-        # overide the lock on any in progress sets
+        # overide the lock, cancel the timer and remove the subscription on any
+        # in progress sets
         self._set_lock.release()
+        self._cb_timer.cancel()
+        self.readback.unsubscribe(self._cid)
         # set the controller to the current value (best option we came up with)
         self.set(self.readback.get())
 
--- a/nslsii/temperature_controllers.py
+++ b/nslsii/temperature_controllers.py
@@ -25,7 +25,6 @@ class Eurotherm(Device):
         self._set_lock = threading.Lock()
 
         # defining these here so that they can be used by `set` and `start`
-        self._status = None
         self._cb_timer = None
         self._cid = None
 
@@ -47,7 +46,7 @@ class Eurotherm(Device):
 
         # define some required values
         set_value = value
-        self._status = DeviceStatus(self)
+        status = DeviceStatus(self)
 
         initial_timestamp = None
 
@@ -62,7 +61,7 @@ class Eurotherm(Device):
                                                           self.timeout.get()))
             self._set_lock.release()
             self.readback.clear_sub(status_indicator)
-            self._status._finished(success=False)
+            status._finished(success=False)
 
         self._cb_timer = threading.Timer(self.timeout.get(), timer_cleanup)
 
@@ -76,7 +75,7 @@ class Eurotherm(Device):
             if abs(value - set_value) < tolerance:
                 if initial_timestamp:
                     if (timestamp - initial_timestamp) > equilibrium_time:
-                        self._status._finished()
+                        status._finished()
                         self._cb_timer.cancel()
                         self._set_lock.release()
                         self.readback.clear_sub(status_indicator)
@@ -92,7 +91,7 @@ class Eurotherm(Device):
         self._cid = self.readback.subscribe(status_indicator)
 
         # hand the status object back to the RE
-        return self._status
+        return status
 
     def stop(self):
         # overide the lock, cancel the timer and remove the subscription on any
--- a/nslsii/temperature_controllers.py
+++ b/nslsii/temperature_controllers.py
@@ -1,4 +1,5 @@
 from ophyd import DeviceStatus, Device, Component as Cpt, EpicsSignal, Signal
+import threading as th
 
 
 class Eurotherm(Device):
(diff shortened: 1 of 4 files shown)


nslsii/temperature_controllers.py
```diff
@@ -1,4 +1,5 @@
 from ophyd import DeviceStatus, Device, Component as Cpt, EpicsSignal, Signal
+import threading as th
```
danielballan: Would you mind avoiding abbreviating imports? The standards are OK (numpy -> np, etc) but this is unnecessarily cryptic.

nslsii/temperature_controllers.py
```diff
@@ -35,29 +37,44 @@ def __init__(self, pv_prefix, **kwargs):
     # define the new set method with the new moving indicator
     def set(self, value):
         # check that a set is not in progress, and if not set the lock.
-        if self._set_lock:
-            raise Exception('attempting to set {}'.format(self.name) +
+        if not self._set_lock.acquire(blocking=False):
+            raise Exception('attempting to set {} '.format(self.name) +
```
danielballan: While we're polishing this, I'll mention a possible improvement that I withheld previously. When a library raises `Exception`, the only way for the call to catch that exception (in order to display a message in a GUI or re-raise a different error, or try again...) is to catch it with `raise Exception` which is overly blunt and might catch many things. Better to be more specific. This could be, for example, a `RuntimError` or -- better yet -- something quite narrow. This design philosophy can be taken too far, of course, but in this case I would suggest:

```
class SetInProgress(RuntimeError):
   ...
```

nslsii/temperature_controllers.py
```diff
@@ -35,29 +37,44 @@ def __init__(self, pv_prefix, **kwargs):
...
+        # setup a cleanup function for the timer, this matches including
+        # timeout in `status` but also ensures that the callback us removed.
+        def timer_cleanup(status):
+            print('Set of {} timed out after {} s'.format(self.name,
+                                                          self.timeout.get()))
+            self._set_lock.release()
+            self.readback.clear_sub(status_indicator)
+            status._finished(success=False)
+
+        cb_timer = th.Timer(self.timeout.get(), timer_cleanup, args=[status])
```
danielballan: The `[status]` jumped out at me. It's a minor thing, but it's safer to use a tuple `(status,)` where possible and basically always for listing positional arguments

nslsii/tests/temperature_controllers_test.py
```diff
@@ -44,6 +45,19 @@ def test_Eurotherm(RE):
...
+        # test that the set will fail after 'timeout'
+        euro.timeout.set(1)
+        with pytest.raises(FailedStatus):
+            RE(mv(euro,100))
+        # ensure callback is removed
+        assert len(euro.readback._callbacks['value']) == 0
+        euro.timeout.set(500) # reset to default for the following tests.
+
+        # test that the lock prevents setting while set in progress
+        with pytest.raises(Exception):
```
danielballan: This is another reason more narrow, case-specific exceptions are good to have. If the test inside that block starts failing for some _other_ reason than the expected one, `Exception` does not distinguish and the test will still pass.

nslsii/temperature_controllers.py
```diff
@@ -68,13 +85,13 @@ def status_indicator(value, timestamp, **kwargs):
...
-        self.readback.subscribe(status_indicator)
+        self._cid = self.readback.subscribe(status_indicator)
 
         # hand the status object back to the RE
         return status
 
     def stop(self):
         # overide the lock on any in progress sets
-        self._set_lock = False
+        self._set_lock.release()
```
danielballan: Should this cancel the timer and clear the subscription?

nslsii/temperature_controllers.py
awalter-bnl: no problem ,this has been changed

nslsii/temperature_controllers.py
awalter-bnl: Also changed

nslsii/tests/temperature_controllers_test.py
awalter-bnl: noted

nslsii/temperature_controllers.py
awalter-bnl: done

nslsii/temperature_controllers.py
```diff
@@ -38,8 +37,8 @@ def __init__(self, pv_prefix, **kwargs):
     def set(self, value):
         # check that a set is not in progress, and if not set the lock.
         if not self._set_lock.acquire(blocking=False):
-            raise Exception('attempting to set {} '.format(self.name) +
-                            'while a set is in progress'.format(self.name))
```
danielballan: Ha, neat. Looks like that with `pytest.raises(Exception)` block in the tests already _was_ masking an error other than the one we were expecting. Good object lesson.

awalter-bnl: not sure that is true, The Travis error is due to Flake8 in .versioneer.py and .ad33.py. 

P.S. I am working on the changes to the stop method (they are a little more involved than first thought).

nslsii/temperature_controllers.py
awalter-bnl: this has been fixed

awalter-bnl:
The Travis error is:

> 1.36s$ flake8 --max-line-length=79
./versioneer.py:941:-13592: W605 invalid escape sequence '\s'
./versioneer.py:941:-13408: W605 invalid escape sequence '\s'
./versioneer.py:941:-13228: W605 invalid escape sequence '\s'
./versioneer.py:941:-11196: W605 invalid escape sequence '\d'
./versioneer.py:941:-8162: W605 invalid escape sequence '\d'
./nslsii/ad33.py:92:11: W605 invalid escape sequence '\d'
The command "flake8 --max-line-length=79" exited with 1.
0.00s$ set -e

These files where not modifed in this PR, and I don't get the same errors when I ru nflake8 locally (I have updated flake8 locally and that did not resolve the issue). Any ideas on why my local flake8 is different to this one?

I will take a look at these on Monday

tacaswell:
This is not how I have seen threading locks normally used.  Rather than the lock it's self being something we check on, it is used to protect the state we care about, ex

```
with self.lock:
    if self._is_running:
        raise
```

We also have a lot of examples of using a locking wrapper on methods.

awalter-bnl:
@tacaswell , thanks for the feedback. For my own learning process is there a reason that using `with self._lock` is better than 'checking' the `self._lock` prior to executing. My understanding is that they are equivalent, with the only major change the fact that I don't have to 'release' the lock at the end.  As there is at least two other 'cleanup' items (i.e stopping the timer and removing the callback) to me I don't see it being a big difference.

So is this just an aesthetic thing or is there something I am missing (either way I am happy to make the change, just want to be clear myself on the reason for future work)?

awalter-bnl:
Thanks @mrakitin for getting Travis to pass !-)

danielballan:
> As there is at least two other 'cleanup' items (i.e stopping the timer and removing the callback) to me I don't see it being a big difference.

Let's think about it. There are two differences I see:

1. It's true that there is more cleanup to do so this doesn't save much/any code, but a context manager provides stronger guarantees that the release _will_ be fun, akin to a `finally` block, which is very important. It also makes it extremely clear to human readers what the scope of the lock is, which is arguably as important.
2. This is a more idiomatic use of a lock. A lock typically _protects_ business-logic state rather than _embodies_ business logic state. "Idomatic" is a bit stronger than "aesthetic": it's not only more elegant, it's easier to for practitioners to interpret and extend (i.e. add more state inside the same lock's context later).

awalter-bnl:
Thanks for the explanation @danielballan, this is helpful.

Unfortuantely, despite several days and numerous searches, I was not able to get the current code to work using a context manager.
The issue as I see it is that the current set method, and the inclusion of the callback that indicates the done status, does not lend itself to the following code layout:

```
 some_lock.acquire()
 try:
   _do something..._
 finally:
   some_lock.release()
```
Which I think is functionally equivalent to (based on what I have found to read online):
```
with some_lock:
    _do something..._
```
When I attempt either of these and run `euro.set('some_value')` the lock is immediately released, at the same time that the `status` object is returned. 

All of the advice I have found on stack overflow, and other sites, regarding how to resolve this issue indicate that the best solution is the one currently in this PR, i.e.
```
if not some_lock.acquire():
    _define callback, ensuring some_lock.release() is called when callback is removed_
    _subscribe callback..._
```
@tacaswell or @danielballan do either of you have any thoughts on a way around this problem so that we can use a context manager with the lock being released after a callback has released itself?

A second, but less critical, issue related to using a context manager is that I would like to run the lock in non-blocking mode so that I can return an exception when trying to set while a set is in process. This can be resolved by creating a 'non-blocking' context manager to wrap around lock however.

danielballan:
There might be a way to do it, but in the interest of moving this forward how about we stick with acquire/release, make the lock protect state rather than embody state, and as you say use non-blocking mode.

awalter-bnl:
either way is fine with me

danielballan:
OK, let's move ahead then.

danielballan:
I think this was accidentally forgotten. @tacaswell, are you content with the current state of this?

awalter-bnl:
Can I get someone (@danielballan, @tacaswell, @mrakitin ?) to take a look at this and merge if it is OK? I think it has been lost in the noise.

nslsii/temperature_controllers.py
```diff
@@ -35,29 +40,44 @@ def __init__(self, pv_prefix, **kwargs):
...
+        status = DeviceStatus(self)
 
         initial_timestamp = None
 
         # grab these values here to avoidmutliple calls.
         equilibrium_time = self.equilibrium_time.get()
         tolerance = self.tolerance.get()
 
+        # setup a cleanup function for the timer, this matches including
+        # timeout in `status` but also ensures that the callback us removed.
```
mrakitin:
```suggestion
        # timeout in `status` but also ensures that the callback is removed.
```

PR #47, opened 2018-10-30
Change count from 5 to 10 in TwoButtonShutter in devices.py

ke-zhang-rd:
This was observed at the XFP experiment, that 5 attempts weren't enough to open the PPS shutter. Decided with @mrakitin to bump it to 10.


danielballan:
The only Travis failure is flake8 on versioneer, which is fixed in another branch.

mrakitin:
Thanks for merging, @danielballan!

Issue #48, opened 2018-12-18
Add loggers aimed at Elasticsearch


danielballan:
We need a Python logging handler that submits data to Elastic. Specifically it should submit at POST request like this:

```
curl -X "PUT" cmb03.cs.nsls2.local:9200/SOME_SENSIBLE_INDEX_NAME/_doc/1 -d '{"hello": "world"}' -H "Content-Type: application/json" 
```

There may already be a nice library of StackOverflow snippet for making HTTP requests from Python loggers. If not, I would just roll something using [requests](http://docs.python-requests.org/en/master/).

mrakitin:
+1 for `requests`. Also, should the authentication be supported?

mrakitin:
In my case I had to add `http://` to the address to have it passed, not refused.

danielballan:
Moving security discussion to a private channel.

mrakitin:
This seems to work:

```py
In [1]: import requests

In [2]: topic = 'test'

In [3]: r = requests.put(f'http://cmb03.cs.nsls2.local:9200/{topic}/_doc/1', json={'hello': 'DAMA tester'})

In [4]: r
Out[4]: <Response [200]>

In [5]: r.text
Out[5]: '{"_index":"test","_type":"_doc","_id":"1","_version":3,"result":"updated","_shards":{"total":2,"successful":1,"failed":0},"_seq_no":2,"_primary_term":1}'
```

danielballan:
Good. As a note to future readers, the URL `http://cmb03.cs.nsls2.local:9200/` should now be `http://elasticsearch.cs.nsls2.local/`. The old one may not work.

Also, refer to this Dropbox Paper for suggested index names: https://paper.dropbox.com/doc/Kafka-Topics--AT8pvSkZTo5zs_yP40M6R5HFAg-Kedt0QGwc0DhH9cZzXkDy

danielballan:
We should wrap the usage demonstrated by [in @mrakitin's comment above](https://github.com/NSLS-II/nslsii/issues/48#issuecomment-448350966) in a [Python logging handler](https://docs.python.org/3/library/logging.html#handler-objects). Something like:

```py
class ElasticHandler(logging.Handler):
    def __init__(self, url):
        self.url = url
        super().__init__()

    def emit(self, record):
        # Extract useful info from the record and put them into a dict.
        response = requests.put(...)
        # Raise an exception if the server we PUT to returns a bad status code.
[...]

PR #49, opened 2019-01-09
MNT: Create loggers.py and add Elasticsearch handler

ke-zhang-rd:
Create loggers for carproto, bluesky and ophyd to Elasticsearch

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,8 +1,4 @@
 [flake8]
-<<<<<<< HEAD
-exclude = .git,__pycache__,build,dist,versioneer.py,nslsii/_version.py,docs/source/conf.py
-=======
 ignore = E402,W504
 exclude = .git,__pycache__,doc/source/conf.py,build,dist,versioneer.py
 max-line-length = 115
->>>>>>> MNT: Add .flake8
--- a/.flake8
+++ b/.flake8
@@ -3,6 +3,6 @@
 exclude = .git,__pycache__,build,dist,versioneer.py,nslsii/_version.py,docs/source/conf.py
 =======
 ignore = E402,W504
-exclude = .git,__pycache__,doc/source/conf.py,build,dist,versioneer.py,caproto/_headers.py,caproto/tests/test_pyepics_compat.py,caproto/_version.py
+exclude = .git,__pycache__,doc/source/conf.py,build,dist,versioneer.py
 max-line-length = 115
 >>>>>>> MNT: Add .flake8
--- a/.flake8
+++ b/.flake8
@@ -1,2 +1,8 @@
 [flake8]
+<<<<<<< HEAD
 exclude = .git,__pycache__,build,dist,versioneer.py,nslsii/_version.py,docs/source/conf.py
+=======
+ignore = E402,W504
+exclude = .git,__pycache__,doc/source/conf.py,build,dist,versioneer.py,caproto/_headers.py,caproto/tests/test_pyepics_compat.py,caproto/_version.py
+max-line-length = 115
+>>>>>>> MNT: Add .flake8
--- a/nslsii/common/ipynb/info.py
+++ b/nslsii/common/ipynb/info.py
@@ -16,11 +16,11 @@ def get_sys_info():
 
     mem = psutil.virtual_memory()
     html += '<tr><td>Total System Memory</td>'
-    html += '<td>{:.4} Mb</td></td>'.format(mem.total/1024**3)
+    html += '<td>{:.4} Mb</td></td>'.format(mem.total / 1024**3)
     html += '<tr><td>Total Memory Used</td>'
-    html += '<td>{:.4} Mb</td></td>'.format(mem.used/1024**3)
+    html += '<td>{:.4} Mb</td></td>'.format(mem.used / 1024**3)
     html += '<tr><td>Total Memory Free</td>'
-    html += '<td>{:.4} Mb</td></td>'.format(mem.free/1024**3)
+    html += '<td>{:.4} Mb</td></td>'.format(mem.free / 1024**3)
 
     html += '<tr><td>Number of CPU Cores</td><td>{}</td></tr>'.format(
         psutil.cpu_count())
--- a/nslsii/loggers.py
+++ b/nslsii/loggers.py
@@ -39,13 +39,17 @@ class ElasticHandler(logging.Handler):
         except Exception:
             self.handleError(record)
 
+def configure_elastic(beamline):
+    url = 'http://elasticsearch.cs.nsls2.local/' + beamline + '.blueksy'
 
-bluesky_logger = logging.getLogger('bluesky')
-carproto_logger = logging.getLogger('carproto')
-ophyd_logger = logging.getLogger('ophyd')
+    bluesky_logger = logging.getLogger('bluesky')
+    bluesky_elastic_hdr = ElasticHandler(url, level = logging.DEBUG)
+    bluesky_logger.addHandler(elastic_hdr)
 
-elastic_hdr = ElasticHandler('http://elasticsearch.cs.nsls2.local/test/_doc/1')
+    carproto_logger = logging.getLogger('carproto')
+    carproto_elastic_hdr = ElasticHandler(url, level = logging.INFO)
+    carproto_logger.addHandler(elastic_hdr)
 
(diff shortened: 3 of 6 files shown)


danielballan:
To test, we need access to a beamline machine (so it can see `elasticsearch.cs.nsls2.local`). I would do the following.

```
ssh xf23id1-srv1
# Create and activate a new conda environment.
conda create -n test-logger python=3.6
conda activate test-logger
# Install the packages we need with pip, including a development version of nslsii.
pip install bluesky ophyd caproto ipython
git clone https://github.com/ke-zhang-rd/nslsii
cd nslsii
git checkout logger
pip install -e .
cd ..
# Start IPython.
ipython
```

Then in IPython:

```py
import nslsii.loggers
from bluesky import RunEngine
RE = RunEngine({})
RE([])  # generates log messages that should go to Elasticsearch
```

Notice that I have _not_ used `bsui` or any IPython profiles, so this will not interfere with any beamline operations.

Next steps:
* We should POST the logs from bluesky, ophyd, and caproto to separate indexes in elastic, replacing `/test/` in the URL with specific names as outlined by Stuart (see https://github.com/NSLS-II/nslsii/issues/48#issuecomment-448363793). This means we'll need a separate `ElasticHandler` instance for each.
* We should think about what log level we want for each library. The DEBUG logs from caproto, for instance, are probably too much information. They would require a lot of storage, and writing them imposes a significant performance overhead. But the DEBUG logs from bluesky are not quite so heavy and _may_ be worth keeping.

danielballan:
Also, we wondered during our Slack call about the difference between `PUT` and `POST`. If you use `PUT`, you must specific a unique ID for that log entry. Thus, `PUT` is idempotent, because PUT-ing it a second time with the same ID is redundent.) If you use `POST`, you leave it up to Elastic to generate a unique ID for you. Thus, `POST` is not idempotent, because POST-ing a second time will generate a second entry with a new, separate unique.

I learned that from this documentation page: https://www.elastic.co/guide/en/elasticsearch/guide/current/index-doc.html

It suggest that you should use PUT if and only if there is a "natural" obvious unique ID. Otherwise, you use POST and let Elastic generate an internal unique ID for you. For these log messages*, I think there is no natural unique ID, so we should use POST. Do you agree?

*There is one possible exception to this statement. The log messages from bluesky that say, "A Document has been emitted with uid=..." _could_ use the document's uid as part of the unique ID for the log message. I am not sure that's the right thing to do here...maybe something to come back to later.

nslsii/loggers.py
```diff
@@ -0,0 +1,52 @@
...
+    url: string
+        url to Elastic server
+
+    level: Logging Levels, optional
+        Handler logging level
+    '''
+    def __init__(self, url, level=logging.INFO):
+        self.url = url
+        self.level = level
+        self.response = None
```
danielballan: This is superfluous now.

ke-zhang-rd: I think url and level are still useful, right?

danielballan: Yes.

nslsii/common/ipynb/info.py
```diff
@@ -16,11 +16,11 @@ def get_sys_info():
...
     mem = psutil.virtual_memory()
     html += '<tr><td>Total System Memory</td>'
-    html += '<td>{:.4} Mb</td></td>'.format(mem.total/1024**3)
+    html += '<td>{:.4} Mb</td></td>'.format(mem.total / 1024**3)
     html += '<tr><td>Total Memory Used</td>'
-    html += '<td>{:.4} Mb</td></td>'.format(mem.used/1024**3)
+    html += '<td>{:.4} Mb</td></td>'.format(mem.used / 1024**3)
     html += '<tr><td>Total Memory Free</td>'
-    html += '<td>{:.4} Mb</td></td>'.format(mem.free/1024**3)
+    html += '<td>{:.4} Mb</td></td>'.format(mem.free / 1024 * 3)
```
mrakitin: Missing second asterisk?

mrakitin (changes requested):
I'd rather separate the style/pep8 checking and the loggers feature, so that the irrelevant files are not touched. But since there are only a few of them, they can be kept. Please notice a critical problem `1000**3` -> `1000 * 3`, which should be fixed. See my comment below.

nslsii/ad33.py
```diff
@@ -89,7 +89,7 @@ class StatsPluginV33(PluginBase):
     Due to https://github.com/areaDetector/ADCore/pull/333
     """
     _default_suffix = 'Stats1:'
-    _suffix_re = 'Stats\d:'
+    _suffix_re = 'Stats\\d:'
```
mrakitin: I'm not sure about this change. Should the `\` be escaped?

mrakitin: They all seem equivalent:
```py
In [1]: _suffix_re = 'Stats\d:'

In [2]: _suffix_re
Out[2]: 'Stats\\d:'

In [3]: _suffix_re2 = 'Stats\\d:'

In [4]: _suffix_re2
Out[4]: 'Stats\\d:'

In [5]: _suffix_re3 = r'Stats\d:'

In [6]: _suffix_re3
Out[6]: 'Stats\\d:'
```
but to make flake8 happy, either 2nd or 3rd option is fine. I'd rather go with the `r'...'` to match the style of recent changes in bluesky.

.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+ignore = E402,W504
+exclude = .git,__pycache__,doc/source/conf.py,build,dist,versioneer.py,caproto/_headers.py,caproto/tests/test_pyepics_compat.py,caproto/_version.py
```
mrakitin: `caproto/_headers.py,caproto/tests/test_pyepics_compat.py,caproto/_version.py` are not relevant. Please remove them.

mrakitin (changes requested):
`caproto/_headers.py,caproto/tests/test_pyepics_compat.py,caproto/_version.py` are not relevant. Please remove them.

mrakitin (approved):

.flake8
```diff
@@ -0,0 +1,4 @@
...
+max-line-length = 115
```
mrakitin: My concern here is that the line length here is 115, but in [.travis.yml](https://github.com/NSLS-II/nslsii/blob/d3f942cda8eac713ac625dbcf4285e108c04f154/.travis.yml#L24) it's set to 79 via command-line arg (which most likely has a priority). I think we should have a single source of truth here, so that only one configuration is controlling the settings in both development environments and in the CI systems.

danielballan: I wonder why it was set to `79` in `.travis.yml` since that is the default anyway. @ke-zhang-rd, I think the right path here is to remove `--max-line-length=79` from [this line](https://github.com/NSLS-II/nslsii/blob/master/.travis.yml#L24) as part of this PR.

danielballan:
The handler itself looks done to me. Next steps:

1. A little thing: we should add `requests` to `requirements.txt`.
2. We should get the index name (currently `'test'`) set to something real. I suggest moving all the code below the `ElasticHandler` definition into a function (`configure_elastic`, perhaps, to follow the pattern set by `configure_base` and `configure_olog`), and make that function expect a beamline (like `'csx'`) as a string. Then the index can be `csx.bluesky` as planned in the Kafka topics document.
3. We need to set the levels of each of our handlers. I think bluesky should be set to DEBUG, and ophyd and caproto should be set to INFO (because DEBUG generates more information than we can afford to store, and because POST-ing all of that data to Elastic would cause data acquisition to slow down significantly).
4. We should then set the levels of all the default handlers, such as `bluesky.current_handler` to `'WARNING'`.

.flake8
mrakitin: I agree with the plan.

danielballan (commented):
This looks good for now. Let's let it sit (i.e. do not merge) until the caproto PR is merged, in case details change, such as the attributes on `record`.

mrakitin:
@jklynch, can it be useful and complimentary to your recent work in #80?
@ke-zhang-rd, what was the blocking issue here?

jklynch:
We decided this is out of scope for #80. I do not want to do any further logging development until we have had time to evaluate the new state of logging on the floor.

mrakitin:
OK, I see.

PR #52, opened 2019-02-19
add a set of generic EPS two state devices

awalter-bnl:
This PR is designed to add some generic EPS two state `ophyd.Device`s (photon shutters, gate valves and pneumatic actuators). As these are common NSLS-II items we would prefer to have a common source for these, instead of the scattered large number we currently have.

### DIFF
--- a/nslsii/EPS_devices/valves_and_shutters.py
+++ b/nslsii/EPS_devices/valves_and_shutters.py
@@ -96,21 +96,21 @@ class EPSTwoStateDevice(Device):
 
     _state1_cmd = FmtCpt(EpicsSignal,
                          '{self.prefix}Cmd:{self._state1_pv_uid}-Cmd',
-                         string=True)
+                         string=True, kind='omitted')
     _state2_cmd = FmtCpt(EpicsSignal,
                          '{self.prefix}Cmd:{self._state2_pv_uid}-Cmd',
-                         string=True)
+                         string=True, kind='omitted')
 
     status = Cpt(EpicsSignalRO, 'Pos-Sts', string=True)
 
-    enabled_status = Cpt(EpicsSignalRO, 'Enbl-Sts', string=True)
+    enabled_status = Cpt(EpicsSignalRO, 'Enbl-Sts', string=True, kind='config')
 
     _fail_to_state1 = FmtCpt(EpicsSignalRO,
                              '{self.prefix}Sts:Fail{self._state1_pv_uid}-Sts',
-                             string=True)
+                             string=True, kind='omitted')
     _fail_to_state2 = FmtCpt(EpicsSignalRO,
                              '{self.prefix}Sts:Fail{self._state2_pv_uid}-Sts',
-                             string=True)
+                             string=True, kind='omitted')
 
     def set(self, val):
         if self._set_st is not None:
@@ -156,36 +156,6 @@ class EPSTwoStateDevice(Device):
         self.status.subscribe(state_cb)
 
         return st
-
-    def stop(self, success):
-        import time
-        prev_st = self._set_st
-        cmd_sig = self._cmd_map[self._stop_str]
-        stop_val = self._target_map[self._stop_str]
-
-        if prev_st is not None:
-            while not prev_st.done:
-                time.sleep(.1)
-        self._was_stopped = (stop_val == self.status.get())
-        st = self.set(cmd_sig)
-        while not st.done:
-            time.sleep(self._retry_sleep_time)
-
-    def resume(self):
-        import time
-        cmd_sig = self._cmd_map[self._resume_str]
-        prev_st = self._set_st
-        if prev_st is not None:
-            while not prev_st.done:
-                time.sleep(.1)
-        if not self._was_stopped:
-            st = self.set(cmd_sig)
-            while not st.done:
-                time.sleep(self._retry_sleep_time)
-
-    def unstage(self):
-        self._was_stopped = True
-        return super().unstage()
 
 
 class PneumaticActuator(EPSTwoStateDevice):
--- a/nslsii/EPS_devices/valves_and_shutters.py
+++ b/nslsii/EPS_devices/valves_and_shutters.py
@@ -9,6 +9,10 @@ class NSLSIIValvesAndShuttersValueError(ValueError):
     pass
 
 
+class NSLSIIValvesAndShuttersEnableError(ValueError):
+    pass
+
+
 class EPSTwoStateDevice(Device):
     '''An ``ophyd.Device`` class for two state EPS objects at NSLS-II.
 
@@ -119,6 +123,12 @@ class EPSTwoStateDevice(Device):
             raise RuntimeError(
                 f'trying to set {self.name} while another set is in progress')
 
+        if self.enabled_status.get()=='False':
+            raise NSLSIIValvesAndShuttersEnableError(
+                f'Attempted to set {self.name} to {val}, but it is disabled '
+                f'check the read only attribute {self.name}.enabled_status '
+                f'to ensure that it is enabled prior to moving.')
+
         cmd_sig = self._cmd_map[val]
         target_val = self._target_map[val]
 
@@ -131,6 +141,7 @@ class EPSTwoStateDevice(Device):
                 st._finished()
                 self._set_st = None
                 self.status.clear_sub(state_cb)
+                cmd_sig._set_thread = None
 
         cmd_enums = cmd_sig.enum_strs
         count = 0
@@ -147,10 +158,10 @@ class EPSTwoStateDevice(Device):
             if value == 'None':
                 if not st.done:
                     time.sleep(self._retry_sleep_time)
-                    cmd_sig.set(1)
+                    cmd_sig.put(1)
                     ts = datetime.datetime.fromtimestamp(
                         timestamp).strftime(self._time_fmtstr)
-                    print(f'** ({ts}) Had to reactuate {self.name} while'
+                    print(f'** ({ts}) Had to reactuate {self.name} while '
                           f'moving to {val}')
                 else:
                     cmd_sig.clear_sub(cmd_retry_cb)
@@ -160,11 +171,6 @@ class EPSTwoStateDevice(Device):
         self.status.subscribe(state_cb)
 
         return st
-
-    def stop(self, success=False):
-        self.status.clear_sub(state_cb)
-        self._set_st._finished(success=success)
-        super().stop()
 
 
 class PneumaticActuator(EPSTwoStateDevice):
--- a/nslsii/EPS_devices/valves_and_shutters.py
+++ b/nslsii/EPS_devices/valves_and_shutters.py
@@ -82,7 +82,7 @@ class EPSTwoStateDevice(Device):
             self._stop_str = self._state2_str
             self._resume_str = self._state1_str
         else:
-            raise nslsiiValves_and_ShuttersValueError(
+            raise NSLSIIValvesAndShuttersValueError(
                 'the kwarg ``stop_str`` in the EpicsTwoStateDevice class '
                 '``__init__`` method needs to be None or needs to match one of '
                 'the kwargs ``state1_str`` or ``state2_str``')
--- a/nslsii/EPS_devices/valves_and_shutters.py
+++ b/nslsii/EPS_devices/valves_and_shutters.py
@@ -53,7 +53,6 @@ class EPSTwoStateDevice(Device):
         The PV associated with moving to 'state 1' is derived from the
         ``prefix`` and the ``state1_pv_uid`` using the following format:
         ``'{prefix}Cmd:{state1_pv_uid}-Cmd'``
-
    '''
 
     def __init__(self, *, prefix, name,
@@ -288,7 +287,6 @@ class GateValve(EPSTwoStateDevice):
         state, the default is 0.5 s.
     **kwargs : dict, optional
         The kwargs passed to the ``ophyd.Device`` ``__init__`` method
-
     '''
 
     def __init__(self, *, prefix, name,
@@ -347,7 +345,6 @@ class PhotonShutter(EPSTwoStateDevice):
         state, the default is 0.5 s.
     **kwargs : dict, optional
         The kwargs passed to the ``ophyd.Device`` ``__init__`` method
-
     '''
 
     def __init__(self, *, prefix, name,
--- a/nslsii/EPS_devices/valves_and_shutters.py
+++ b/nslsii/EPS_devices/valves_and_shutters.py
@@ -159,6 +159,10 @@ class EPSTwoStateDevice(Device):
 
         return st
 
+    def stop():
+        self.status.clear_sub(state_cb)
+        self._set_st._finished(succsess=False)
+        super().stop()
 
 class PneumaticActuator(EPSTwoStateDevice):
     '''An ``ophyd.Device`` for EPS driven pneumatic actuators at NSLS-II.
--- a/nslsii/EPS_devices/valves_and_shutters.py
+++ b/nslsii/EPS_devices/valves_and_shutters.py
@@ -126,7 +126,7 @@ class EPSTwoStateDevice(Device):
         def state_cb(value, timestamp, **kwargs):
             value = enums[int(value)]
             if value == target_val:
-                self._set_st._finished()
+                st._finished()
                 self._set_st = None
                 self.status.clear_sub(state_cb)
 
@@ -139,6 +139,8 @@ class EPSTwoStateDevice(Device):
             count += 1
             if count > self._num_retries:
                 cmd_sig.clear_sub(cmd_retry_cb)
+                self._set_st = None
+                self.status.clear_sub(state_cb)
                 st._finished(success=False)
             if value == 'None':
                 if not st.done:
--- a/nslsii/EPS_devices/valves_and_shutters.py
+++ b/nslsii/EPS_devices/valves_and_shutters.py
@@ -62,7 +62,10 @@ class EPSTwoStateDevice(Device):
                  num_retries=1, retry_sleep_time=0.5,
                  stop_str=None, **kwargs):
 
-        super().__init__(prefix, name, **kwargs)
+        self._state1_pv_uid = state1_pv_uid
+        self._state2_pv_uid = state2_pv_uid
+
+        super().__init__(prefix=prefix, name=name, **kwargs)
 
         self._set_st = None
         self.read_attrs = ['status']
@@ -70,8 +73,6 @@ class EPSTwoStateDevice(Device):
         self._state2_val = state2_val
         self._state1_str = state1_str
         self._state2_str = state2_str
-        self._state1_pv_uid = state1_pv_uid
-        self._state2_pv_uid = state2_pv_uid
         self._num_retries = num_retries
         self._retry_sleep_time = retry_sleep_time
         if stop_str == state1_str:
--- a/nslsii/EPS_devices/valves_and_shutters.py
+++ b/nslsii/EPS_devices/valves_and_shutters.py
@@ -123,7 +123,7 @@ class EPSTwoStateDevice(Device):
             raise RuntimeError(
                 f'trying to set {self.name} while another set is in progress')
 
-        if self.enabled_status.get()=='False':
+        if self.enabled_status.get() == 'False':
             raise NSLSIIValvesAndShuttersEnableError(
                 f'Attempted to set {self.name} to {val}, but it is disabled '
                 f'check the read only attribute {self.name}.enabled_status '
--- a/nslsii/EPS_devices/valves_and_shutters.py
+++ b/nslsii/EPS_devices/valves_and_shutters.py
@@ -164,9 +164,9 @@ class EPSTwoStateDevice(Device):
         self._set_st._finished(success=False)
         super().stop()
 
+
 class PneumaticActuator(EPSTwoStateDevice):
     '''An ``ophyd.Device`` for EPS driven pneumatic actuators at NSLS-II.
-
 
     This is for use with NSLS-II EPS based pneumatic actuators that have two
     distinct states 'In' and 'Out' and are actuated into these two states by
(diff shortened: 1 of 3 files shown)


prjemian:
Just gone through a revision of our shutter infrastructure with the goal of making it uniform regardless of the hardware implementation.  For any shutter, `shutter.open()`, `shutter.close()` are viable commands.  The code relies on (hardware-dependent) subclass implementation of a `state` method that tells if *open*, *close*, or *unknown*.  Also, any shutter can be used in a plan such as:  `yield from bps.mv(shutter, "open")`.  There are properties for `shutter.isOpen` and `shutter.isClosed` and some other useful stuff.

You might take a look at the source code:

* [ShutterBase](https://github.com/BCDA-APS/apstools/blob/master/apstools/devices.py#L238)
* [OneSignalShutter](https://github.com/BCDA-APS/apstools/blob/master/apstools/devices.py#L423)
* [EpicsMotorShutter](https://github.com/BCDA-APS/apstools/blob/master/apstools/devices.py#L1102)
* [EpicsOnOffShutter](https://github.com/BCDA-APS/apstools/blob/master/apstools/devices.py#L1159)
* [ApsPssShutter](https://github.com/BCDA-APS/apstools/blob/master/apstools/devices.py#L517)
* [ApsPssShutterWithStatus](https://github.com/BCDA-APS/apstools/blob/master/apstools/devices.py#L609)
* [SimulatedApsPssShutterWithStatus](https://github.com/BCDA-APS/apstools/blob/master/apstools/devices.py#L757)


The shutter simulator is a great tool.  We use it like this in our ipython startups so we can test our control software during maintenance periods:

```
if aps.inOperations:
   safety_shutter = ApsPssShutterWithStatus("pv_prefix", name="safety_shutter")
else:
   print("!!!!! Note: using simulated safety_shutter !!!!")
   safety_shutter = SimulatedApsPssShutterWithStatus(name="safety_shutter")
```

prjemian:
@awalter-bnl : You correct.  Operation of a valve looks like a shutter.  That could be subclassed from [ShutterBase](https://github.com/BCDA-APS/apstools/blob/master/apstools/devices.py#L238) or possibly used directly from [EpicsOnOffShutter](https://github.com/BCDA-APS/apstools/blob/master/apstools/devices.py#L1159)

awalter-bnl:
Thanks for the feedback @prjemian I will take a look at your shutter implementation, and see what is there. This is actually heavily based on exisiting code we have which is seperated between beamlline profiles at NSLS-II at the moment and we are attempting to `centralize` it.

Unfortunately I am not yet convinced that sharing the code for facility dependent shutters is a good idea because:
1. our shutters are quite anachronistic in that they sometimes need to be 'poked' multiple times to actually change state so we have some built-in mechanisms to take care of that. 
2. our shutters/ gate-valves and pneumatic actuators all share a similar IOC (based on our EPS systems plc controls) and so I think something facility specific which we can tune to this use case is better for us.

What do others think?

tacaswell:
One thing that I think should be given a bit of thought here is the error handling around trying to start a while another set is in motion.  Currently, the status object stashed on the shutter object is cleared on the success of the motion, but if you tell it to open and, for what ever reason, in never actually opens the status object will never be cleared and we won't be able to use the shutter again without restarting bluesky (or reaching in and clearing it by hand).  This should probably be cleaned up a bit, but I am not sure the best way.  If I recall correctly, the original version of this class was written while a beamline scientist was watching me so I took the "fail hard, sort it out later" approach.

I think our options here are:
 - add logic else where to make sure we clear this, but leave the hard fail
 - unconditionally fail the previous move like `Positioner` does [1].  I am not sure that this will work with all of the devices that this will control as the IOC can not be interrupted cleanly / will not take new direction while it is actively moving.  I am also not sure that it will provide clear feedback that it ignored you
 - as we only have 2 states, detect if we are going to the same place as we were just told to and
   simply return the old status objects, if going the other way do one of the first two.

[1] https://github.com/NSLS-II/ophyd/blob/204c8b8bfb3d907670c2a083b7f5c2dbe885c7f5/ophyd/positioner.py#L185

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,363 @@
...
+                else:
+                    cmd_sig.clear_sub(cmd_retry_cb)
+
+        cmd_sig.subscribe(cmd_retry_cb, run=False)
+        cmd_sig.set(1)
+        self.status.subscribe(state_cb)
+
+        return st
+
+    def stop(self, success):
```
tacaswell: I am not sure that we want to have this as the base-class `stop` logic.  This means that `%mv shutter 'Open'` will 'fail' because it will open and then on the way out the RE will call 'stop' on it which will re-close it. Same comment on 'resume' below.  

This logic is so that when stopping / resuming a scan at XFP we would be sure that the front end shutter would close and stop cooking their sample.  We should check that we do not also need to put this logic in `pause` as I do not recall of the top of my head if RE calls `stop` when suspenders fire (I think it does...).

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,363 @@
...
+                'the kwargs ``state1_str`` or ``state2_str``')
+
+        self._cmd_map = {self._state1_str: self._state1_cmd,
+                         self._state2_str: self._state2_cmd}
+        self._target_map = {self._state1_str: self._state1_val,
+                            self._state2_str: self._state2_val}
+
+        self._time_fmtstr = '%Y-%m-%d %H:%M:%S'
+
+    _state1_cmd = FmtCpt(EpicsSignal,
```
tacaswell: can you put a default kind on everything?  I suspect that `status` should be 'normal', `enabled_status` should be 'configuration' and everything else should be omitted.

the `_fail_to_state` may also want to be normal?

awalter-bnl:
>add logic else where to make sure we clear this, but leave the hard fail

I am not sure where this logic will live, or what it will look like but it seems like the most likely of the @tacaswell 's 3 options to work.

> unconditionally fail the previous move like Positioner does [1]. I am not sure that this will work with all of the devices that this will control as the IOC can not be interrupted cleanly / will not take new direction while it is actively moving. I am also not sure that it will provide clear feedback that it ignored you

I suspect @tacaswell is correct here, in that although they share an IOC template each of these cases has very different hardware.

> as we only have 2 states, detect if we are going to the same place as we were just told to and
simply return the old status objects, if going the other way do one of the first two.

This may also work as a solution but the second part causes problems. For instacne there is no way for any of these item to be 'stopped' mid-way between opening and closing (as they are all pneumatic actuators). So if we are not going in the same direction as we previously asked we _must_ wait until that move is done before requesting the next move.

nslsii/EPS_devices/valves_and_shutters.py
awalter-bnl: Yes this was an issue that @mrakitin mentioned for existing shutters at CMS as well last week. I have made this change. FYI It does call 'stop' when suspenders fire, as the guys at SIX asked me 2 weeks ago what was the logic for a gate-valve closing when the suspender 'fires'. Once I told them they where actually happy that it did it !-).

I have removed this logic.

awalter-bnl: resolved

nslsii/EPS_devices/valves_and_shutters.py
awalter-bnl: done

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,333 @@
...
+
+        st = self._set_st = DeviceStatus(self)
+        enums = self.status.enum_strs
+
+        def state_cb(value, timestamp, **kwargs):
+            value = enums[int(value)]
+            if value == target_val:
+                self._set_st._finished()
+                self._set_st = None
+                self.status.clear_sub(state_cb)
```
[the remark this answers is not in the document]
awalter-bnl: Now that I look in more detail at the example you give from nslsii I found that line 45 in nslsii/nslsii/devicces.py sets st to be self._set_st, so I think that the two approaches are then equivalent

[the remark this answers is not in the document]
tacaswell: There are some subtle thread-related race conditions here, will leave longer comments in the main thread so it does not get lost.  This is one place where the difference between the versions needs to be carefully reconciled.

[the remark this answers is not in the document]
mrakitin: @awalter-bnl, let's wait for @tacaswell's comments. I remember his in-person explanation, but don't remember every detail.

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,333 @@
...
+        cmd_enums = cmd_sig.enum_strs
+        count = 0
+
+        def cmd_retry_cb(value, timestamp, **kwargs):
+            nonlocal count
+            value = cmd_enums[int(value)]
+            count += 1
+            if count > self._num_retries:
+                cmd_sig.clear_sub(cmd_retry_cb)
+                st._finished(success=False)
```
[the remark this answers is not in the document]
mrakitin: I think at this point `st` and `self._set_st` are different: the former is a status object, and the latter is `None`.

It can be demonstrated by this snippet:
```py
In [1]: from ophyd.status import Status

In [2]: st1 = st2 = Status()

In [3]: st1
Out[3]: Status(obj=None, done=False, success=False)

In [4]: st2
Out[4]: Status(obj=None, done=False, success=False)

In [5]: st1 = None

In [6]: st1

In [7]: st2
Out[7]: Status(obj=None, done=False, success=False)
```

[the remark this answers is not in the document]
awalter-bnl: thanks for the snippet @mrakitin this was useful for me to see the difference. I will leave off making any changes so that I can see @tacaswell's detailed discussion on race conditions

tacaswell:
There are three places we need to keep a reference to the `DeviceStatus` object on the this device

 1. on the instance its self to keep track off if there is current a set in progress
 2. the returned reference so the ultimate caller can wait for the device to finish
 3. in the callback that gets handed to the status object so it can notify the rest of the world when it is done.

The epics callback runs on a thread (actually a c-thread that libca owns) so we need to worry about concurrency between the various parts.

If we have the simplest verison of this code it looks something like

```python
def set():
    if self._st is not None:
        raise

    self._st = Status()

    def cb(value, **kwargs):
       if value:
           self._st._finished()
           self._st = None
           self.pv.remove_cb(cb)

   self.pv.subscribe(cb)
   return self._st
```

which will work _most_ of the time, but there are a few gotcha's in the timing.  If the callback fires _before_ the return statement is executed, `self._st` will be `None` which will break the RE (as it expects a `Status` object to be returned).  Another way this can fail is say you have the plan

```python
yield from bps.mov(shutter,' 'Open')
yield from bps.mov(shutter, 'Close')
```

you can get the timing where

 1. the open callback sets the status objects as done
 2. the wait in the RE plan returns, the next message is processed and we try to set the shutter colsed
 3. that set blows up because the callback has not set `self._st` to None

I third race condition is that if the callback fires a second time before the first one finishes (which I don't _think_ should be possible, but that depends on the details of how both libca and ophyd deal with dispatching the callbacks bubbling out of the monitors) where you could hit the `self._st = None` from the first callback before you hit the `self._st._finished()` from the second.

Both of these timings will be relatively rare. 

The following is a bit more defensive

```python
def set():
    if self._st is not None:
        raise

    st = self._st = Status()

    def cb(value, **kwargs):
       if value:
           self._st = None
           st._finished()
           self.pv.remove_cb(cb)

   self.pv.subscribe(cb)
   return st
```

by keeping it's own reference to the `Status` object (both to return and to have the callback close over) makes it robust to all three of the race conditions above.

tacaswell:
We also need to make sure that calling `stop` on the device correctly cleans it's self up (failing the status object, pulling the callback off, clearing things so the device can be moved again).

It is possible that we should be using threading locks here (rather than relying on careful ordering) and that the in-line callback is not the correct choice to make sure we can reliably remove the callback (by subscribing a method) or we should be tracking the id returned by subscribe rather than using the callable.

nslsii/EPS_devices/valves_and_shutters.py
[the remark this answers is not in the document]
awalter-bnl: I have resolved this, see response to @tacaswell below

nslsii/EPS_devices/valves_and_shutters.py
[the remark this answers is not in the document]
awalter-bnl: I have resolved this, see response to @tacaswell below

awalter-bnl:
Thanks for the detailed response @tacaswell , and for picking up on this initially @mrakitin. I have made the changes as requested by @mrakitin and added the custom `stop` method as requested by @tacaswell .

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+
+        cmd_sig.subscribe(cmd_retry_cb, run=False)
+        cmd_sig.set(1)
+        self.status.subscribe(state_cb)
+
+        return st
+
+    def stop():
+        self.status.clear_sub(state_cb)
+        self._set_st._finished(succsess=False)
```
mrakitin:
```suggestion
        self._set_st._finished(success=False)
```

mrakitin: If I remember correctly, the stop method requires `success` keyword, which should be passed to the line 164 (see my suggestion above).

awalter-bnl: @mrakitin, good catch, I have made the change

mrakitin: Thanks @awalter-bnl! I think the `success` kwarg should still be added, as it's done in https://github.com/NSLS-II/nslsii/blob/0ff403eb24b5a1acb79f93e434a7415b06fefb2b/nslsii/devices.py#L88

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+        self._num_retries = num_retries
+        self._retry_sleep_time = retry_sleep_time
+        if stop_str == state1_str:
+            self._stop_str = self._state1_str
+            self._resume_str = self._state2_str
+        elif stop_str == state2_str or stop_str is None:
+            self._stop_str = self._state2_str
+            self._resume_str = self._state1_str
+        else:
+            raise nslsiiValves_and_ShuttersValueError(
```
mrakitin: Sorry I didn't change it here as well with my original suggestion.
```suggestion
            raise NSLSIIValvesAndShuttersValueError(
```

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+        cmd_sig.set(1)
+        self.status.subscribe(state_cb)
+
+        return st
+
+    def stop():
+        self.status.clear_sub(state_cb)
+        self._set_st._finished(success=False)
+        super().stop()
+
```
mrakitin:
```suggestion


```

mrakitin: Should we also add a `resume` method as it's done here? https://github.com/NSLS-II/nslsii/blob/0ff403eb24b5a1acb79f93e434a7415b06fefb2b/nslsii/devices.py#L99-L108

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+class PneumaticActuator(EPSTwoStateDevice):
+    '''An ``ophyd.Device`` for EPS driven pneumatic actuators at NSLS-II.
+
```
mrakitin:
```suggestion
```

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+    ``RE(mv(my_pneumatic_actuator, 'out')``
+
+    Parameters
+    ----------
+    prefix : str, keyword only
+        The PV prefix for all components of the device
+    name : str, keyword only
+        The name of the device as will be reported via read()
+    state1_val : float, int or str, optional
+        The value displayed when the device is read that signifys state 1
```
mrakitin:
```suggestion
        The value displayed when the device is read that signifies state 1
```

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+    state2_val : float, int or str, optional
+        The value displayed when the device is read that signifys state 2
```
mrakitin:
```suggestion
        The value displayed when the device is read that signifies state 2
```

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+        The kwargs passed to the ``ophyd.Device`` ``__init__`` method
+    '''
+
+    def __init__(self, *, prefix, name,
+                 state1_val='In', state2_val='Out',
+                 state1_str='in', state2_str='out',
+                 num_retries=1, retry_sleep_time=0.5,
+                 stop_str=None, **kwargs):
+
+        kwargs = dict(state1_pv_uid='In', state2_pv_uid='Out',
```
mrakitin: ~~Why can't we use the `state1_val` and `state2_val` instead of `In` and `Out` respectively?~~
I saw another use like this in https://github.com/NSLS-II/nslsii/pull/52/files#diff-e886eb2e4c555c0292a7ed27bc136211R275, disregard my comment.

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+                      state1_val=state1_val, state2_val=state2_val,
+                      state1_str=state1_str, state2_str=state2_str,
+                      num_retries=num_retries,
+                      retry_sleep_time=retry_sleep_time,
+                      stop_str=stop_str, **kwargs)
+        super().__init__(prefix, name, **kwargs)
+
+
+class GateValve(EPSTwoStateDevice):
+    '''An ``ophyd.Device`` for EPS driven gate valves at NSLS-II
```
mrakitin:
```suggestion
    '''An ``ophyd.Device`` for EPS driven gate valves at NSLS-II.
```

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+    ``RE(mv(my_gate_valve, 'close')``
+
+    Parameters
+    ----------
+    prefix : str, keyword only
+        The PV prefix for all components of the device
+    name : str, keyword only
+        The name of the device as will be reported via read()
+    state1_val : float, int or str, optional
+        The value displayed when the device is read that signifys state 1
```
mrakitin:
```suggestion
        The value displayed when the device is read that signifies state 1
```

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+    state2_val : float, int or str, optional
+        The value displayed when the device is read that signifys state 2
```
mrakitin:
```suggestion
        The value displayed when the device is read that signifies state 2
```

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+    ``RE(mv(my_shutter, 'close')``
+
+    Parameters
+    ----------
+    prefix : str, keyword only
+        The PV prefix for all components of the device
+    name : str, keyword only
+        The name of the device as will be reported via read()
+    state1_val : float, int or str, optional
+        The value displayed when the device is read that signifys state 1
```
mrakitin:
```suggestion
        The value displayed when the device is read that signifies state 1
```

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+    state2_val : float, int or str, optional
+        The value displayed when the device is read that signifys state 2
```
mrakitin:
```suggestion
        The value displayed when the device is read that signifies state 2
```

nslsii/EPS_devices/valves_and_shutters.py
```diff
@@ -0,0 +1,339 @@
...
+    state1_str : str, optional
+        The string value to be passed to ``RE(mv(device, val))`` to move to
+        state 1
+    state2_str : str, optional
+        The string value to be passed to ``RE(mv(device, val))`` to move to
+        state 2
+    num_retries : int, optional
+        This number of attempts at changing the state prior to raising an
+        error, the default is 5.
```
mrakitin: Let's put 10 as it was previously done via https://github.com/NSLS-II/nslsii/pull/47.
```suggestion
        error, the default is 10.
```

nslsii/devices.py
```diff
@@ -74,7 +74,7 @@ def cmd_retry_cb(value, timestamp, **kwargs):
                         .strftime(_time_fmtstr)
                     if count > 2:
                         msg = '** ({}) Had to reactuate shutter while {}ing'
-                        print(msg.format(ts, val if val is not 'Close'
+                        print(msg.format(ts, val if val != 'Close'
```
mrakitin: 👍

nslsii/EPS_devices/valves_and_shutters.py
awalter-bnl: thanks @mrakitin I have made this change in an upcoming commit

awalter-bnl:
thanks for the review @danielballan

awalter-bnl:
latest commit adds tests againsts the new caproto sim IOC (thanks @malitsky for the work on that !-) ).

awalter-bnl:
@mrakitin can you take another pass over this? I think it is ready

awalter-bnl:
I am guessing this is dead due to the lack of responses, if others feel differently feel free to reopen.

awalter-bnl:
re-opening at the request of @mrakitin

PR #53, opened 2019-02-20
add a collection of generic slit devices

awalter-bnl:
This PR is designed to add some generic slit devices based on `ophyd.Devices` (FourBladeSlits, ScanAndApertureBaffleSlits and FrontEndSlits). As these are common NSLS-II items we would prefer to have a common source for these, instead of the scattered large number we currently have.

### DIFF
--- a/nslsii/motor_devices/slits.py
+++ b/nslsii/motor_devices/slits.py
@@ -19,9 +19,9 @@ def slit(name='', axes={'hg': '-Ax:HG}Mtr', 'hc': '-Ax:HC}Mtr',
 
     ..code ::
 
-        HorizSlit=BaffleSlit(name='Horiz_Slit',
-                             axes={'hg': '-Ax:HG}Mtr', 'hc':'-Ax:HC}Mtr'})
-        my_slit=HorizSlit(PV_prefix, name='my_slit')
+        HorizSlit = slit(name='Horiz_Slit',
+                         axes={'hg': '-Ax:HG}Mtr', 'hc':'-Ax:HC}Mtr'})
+        my_slit = HorizSlit(PV_prefix, name='my_slit')
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


awalter-bnl:
based on feedback regarding the lakeshore controller PR ( #54 ) I have converted the base class into a factory function for generating classes.

awalter-bnl:
The errors are caused by a dependency on an unreleased version of Ophyd. I will resolve that as soon as we have a release.

nslsii/motor_devices/slits.py
```diff
@@ -0,0 +1,151 @@
+from ophyd import (Device, Component, EpicsMotor, PVPositioner, EpicsSignal,
+                   EpicsSignalRO)
+from ophyd.device import create_device_from_components
+
+
+def slit(name='', axes={'hg': '-Ax:HG}Mtr', 'hc': '-Ax:HC}Mtr',
```
tacaswell: mutable default values are technically possible, but are very bad practice.   It is better to default to `None` and have this dictionary in the body.  That also gives a nice path to overriding just one of the names as you can do

```
dflt = {}
dflt.update(axes)
```

tacaswell: do we want to make this function private?

Reading down a bit more, no-default maybe better here.

I also think that we should not have a default name and force the callers of this function to set a class name.

awalter-bnl: I have made the default `None`, and added an `if default is None:` section to set the default.

I am against having no-default, as having a 'default' helps with the social engineering of moving everyone to a common solution, but provides the flexibility o have custom cases. !-).

I have made `name` non-optional.

awalter-bnl:
@danielballan thanks

awalter-bnl:
cycled the PR close/open to restart the tests.

awalter-bnl:
am guessing that judging by the fact this hasn't been merged that it is no longer required. closing it out, feel free to resurrect it if it is needed in the future.

awalter-bnl:
reopening at the request of @mrakitin

PR #54, opened 2019-02-20
add a generic Lakeshore model 336 device

awalter-bnl:
This PR is designed to add a generic Lakeshore model 336 device based on `ophyd.Device`. As these are standard NSLS-II items we would prefer to have a common source for these, instead of the scattered large number we currently have. I am yet to determine if we are using a pulbically available IOC, if it turns out we are I will instead move this device to `ophyd` instead of this repo.


nslsii/temperature_controllers/lakeshore.py
```diff
@@ -0,0 +1,126 @@
...
+                                       {'write_pv': '}Out:MaxI-SP'}),
+                       'resistance': (EpicsSignalRO, '}Out:R-RB',
+                                      {'write_pv': '}Out:R-SP'})}
+        output = DynamicDeviceComponent(output_dict)
+
+    def __init__(self, *args, temperatures=['A', 'B', 'C', 'D'],
+                 controls=[1, 2, 3, 4], **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.temperature = DynamicDeviceComponent(
```
tacaswell: These should be defined at the class level, not in the `__init__`.  This should look something like https://github.com/NSLS-II/ophyd/blob/fe812226a193adab0515a81e39fb5e5c22d2ea0c/ophyd/scaler.py#L24-L60

Because there are not defined at class instantiation time, the logic in https://github.com/NSLS-II/ophyd/blob/e098e7f873c8a564fc4c66d2f512286aba3f51cf/ophyd/device.py#L831-L834 will be triggered and correclty set up.  If you do `obj.temperature` you are going to get back the DDC instance, not and instance of `_Temperature`.   This is essentially:

```python
In [18]: class Foo:
    ...:     def __init__(self):
    ...:         self.a = property(lambda self, x: 5, lambda self, x, v: None)

In [19]: f = Foo()

In [20]: f.a
Out[20]: <property at 0x7f75f494b2c8>

```

which does not do what we want.

nslsii/temperature_controllers/lakeshore.py
```diff
@@ -0,0 +1,126 @@
...
+        A list of temperature sensor 'channels' to instantiate, the default is
+        ['A','B','C','D'].
+    controls: list
+        A list of heater control 'channels' to instantiate, the default is
+        [1,2,3,4].
+    **kwargs: dict
+        The keyword arguments passed to ``ophyd.Device``.
+    '''
+
+    def _set_fields(fields, device, prefix, **kwargs):
```
tacaswell: While not wrong, this naming is very confusing.  When passing in a class (that is an instance of `type`) you usually name it `cls` or `klass` (but not `class` because that is a reserved keyword) to be clear to the user that this is not an _instance_ of the class, but the class it's self.

nslsii/temperature_controllers/lakeshore.py
```diff
@@ -0,0 +1,126 @@
...
+        '''A function that allows for the components to be dynamically set.
+
+        Parameters
+        ----------
+        fields: list
+            A list of field identifiers to include, for ``_Temperature``
+            components an example is ['A','B','C','D'] and for ``_Control``
+            components an example is [1,2,3,4].
+        device: object
```
tacaswell: Similar to above, this is not technically wrong (`type` instances are a sub-class of `object`) but this suggests you want _instances_ of the device, not the class of the device.

nslsii/temperature_controllers/lakeshore.py
tacaswell: This also won't work as python will interpret this as a method, bind it to the instance and when you use it as `__init__` I think that this will raise a type error because you passed it the wrong number of positional args.

You ether need to add a `@staticmethod` decorator or just move it out of the class.

nslsii/temperature_controllers/lakeshore.py
```diff
@@ -0,0 +1,126 @@
...
+        **kwargs: dict
+            The kwargs that are to be passed into the call to ``device``.
+        '''
+        out_dict = OrderedDict()
+        for field in fields:
+            suffix = f'{prefix}{field}'
+            out_dict[f'{field}'] = (device, suffix, kwargs)
+        return out_dict
+
+    class _Temperature(Device):
```
tacaswell: I am not sure defining the classes in the class helps with the clairity.  Making them private is good though.

nslsii/temperature_controllers/lakeshore.py
```diff
@@ -0,0 +1,126 @@
...
+        readback = Component(EpicsSignalRO, '}T-RB')
+        done = Component(EpicsSignalRO, '}Sts:Ramp-Sts')
+        # top level attributes
+        heater_range = Component(EpicsSignal, '}Val:Range-Sel')
+        heater_status = Component(EpicsSignalRO, '}Err:Htr-Sts')
+        mode = Component(EpicsSignal, '}Mode-Sel')
+        enable = Component(EpicsSignal, '}Enbl-Sel')
+        target_channel = Component(EpicsSignal, '}Out-Sel')
+        # ramp attributes
+        ramp_dict = {'enabled': (EpicsSignal, '}Enbl:Ramp-Sel'),
```
tacaswell: This is going to be available on the final class, probbaly better to define this in-line with the DDC call it is used in.

nslsii/temperature_controllers/lakeshore.py
```diff
@@ -0,0 +1,126 @@
...
+        # PID loop parameters
+        pid_dict = {'proportional': (EpicsSignal, '}Gain:P-RB',
+                                     {'write_pv': '}Gain:P-SP'}),
+                    'integral': (EpicsSignal, '}Gain:I-RB',
+                                 {'write_pv': '}Gain:I-SP'}),
+                    'derivative': (EpicsSignal, '}Gain:D-RB',
+                                   {'write_pv': '}Gain:D-SP'})}
+        pid = DynamicDeviceComponent(pid_dict)
+        # output parameters
+        output_dict = {'current': (EpicsSignal, '}Out-I'),
```
tacaswell: same comment and `ramp_dict` above.

nslsii/temperature_controllers/lakeshore.py
awalter-bnl: Thanks for the feedback @tacaswell , I suspected strongly that what I was doing would not work and added it in order to start a conversation !-). 

While I agree that this does not do what we want, the common approach you suggest also does not do what we want. After looking at the various instances of this class that exist on the floor at present I found that the major difference between each instance is that they have different numbers of 'sensor' and 'control' attributes. Some beamlines only use 2 sensors and 1 control, others use all 4 sensors and 2 controls etc, etc. In fact there are a few locations where they use a variant of this model that has 8 sensors and 4 controls in the one unit.

As I think it does not make sense to force everyone to include extra sensors and controls that they don't have, or have nothing connected too, there is no way to come up with a 'standard' that is flexible enough to account for the current use cases unless we have a way of defining the number of sensors and controls attributes at instantiation time. Is there an 'acceptable' way that will work to do this that anyone can think of?

nslsii/temperature_controllers/lakeshore.py
awalter-bnl: I have changed  `device` to `cls` as suggested, see comment above about moving this from __init__.

nslsii/temperature_controllers/lakeshore.py
awalter-bnl: resolved

nslsii/temperature_controllers/lakeshore.py
awalter-bnl: I think it is clearer doing it this way, but could be convinced to change it if the consensus is not with me.

nslsii/temperature_controllers/lakeshore.py
awalter-bnl: good call, done

nslsii/temperature_controllers/lakeshore.py
awalter-bnl: also done

nslsii/temperature_controllers/lakeshore.py
awalter-bnl: have made this a factory function for classes which solves this issue

nslsii/temperature_controllers/lakeshore.py
awalter-bnl: have made this a factory function which solves this issue

nslsii/temperature_controllers/lakeshore.py
awalter-bnl: have made this a factory function which solves this issue

awalter-bnl:
After a slack discussion I have converted this from a _generic_ class to _factory function_ that generates classes.

awalter-bnl:
The errors are caused by a dependency on an unreleased version of Ophyd. I will resolve that as soon as we have a release.

awalter-bnl:
am guessing that this is a dead PR as it has gone stale. am closing it out.

awalter-bnl:
reopening at the request of @mrakitin

PR #59, opened 2019-04-28
EPSTwoStateIOC

malitsky:


nslsii/iocs/tests/test_epstwostate_ioc.py
```diff
@@ -1,6 +1,7 @@
-# import subprocess
-# import os
-# import sys
+import subprocess
```
awalter-bnl: In general we like to have these imports in alphabetical order, can you move this down one line?

nslsii/devices.py
```diff
@@ -74,7 +74,7 @@ def cmd_retry_cb(value, timestamp, **kwargs):
                         .strftime(_time_fmtstr)
                     if count > 2:
                         msg = '** ({}) Had to reactuate shutter while {}ing'
-                        print(msg.format(ts, val if val is not 'Close'
+                        print(msg.format(ts, val if val != 'Close'
```
awalter-bnl: good catch, thanks for putting this change in.

nslsii/iocs/eps_two_state_ioc_sim.py
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python3
+from caproto.server import pvproperty, PVGroup
+from caproto.server import template_arg_parser, run
+from caproto import ChannelType
+
+
+class EPSTwoStateIOC(PVGroup):
+    """
+    Simulates multiple-attempt issue for two-state device.
```
awalter-bnl:
```suggestion
    Simulates an EPS Two State device including multiple-attempt issue for two-state device.
    
    This IOC is used to simulate an EPS Two State Device, for testing or development purposes. Known EPS two state devices include Photon shutters, gate valves and Pneumatic actuators at NSLS-II. It simulates known issues with some of these including: A hardware error (when some attempts at sending the command fail and we need to 'kick' the device a few times to get it to actuate), A position status error (it sometimes does not reach the final state but remains 'between states') and an enable state change error (there is an 'enable' PV controlled by the control room that determines if the device can be operated or not). A parameter (described below) allows each of these error paths to be tested against
```

I prefer longer doc-strings as these are used in our documentation as well.

nslsii/iocs/tests/test_epstwostate_ioc.py
```diff
@@ -0,0 +1,99 @@
...
+
+    opn_cmd = FmtCpt(EpicsSignal,
+                     '{self.prefix}Cmd:Opn-Cmd',
+                     string=True, kind='omitted')
+    cls_cmd = FmtCpt(EpicsSignal,
+                     '{self.prefix}Cmd:Cls-Cmd',
+                     string=True, kind='omitted')
+
+
+def test_epstwostate_ioc():
```
awalter-bnl: We should probably add tests of the other two 'failure modes', or include them as a part of this test.

malitsky:
@awalter-bnl , thank you for the comments. The PR has been updated with the description and new tests.

PR #62, opened 2019-05-06
EpicsMotorIOC

malitsky:


nslsii/iocs/epics_motor_ioc_sim.py
```diff
@@ -0,0 +1,148 @@
...
+                              enum_strings=_cmd_states,
+                              dtype=ChannelType.ENUM,
+                              name='.HOMR')
+
+    # Methods
+
+    @user_setpoint.putter
+    async def user_setpoint(self, instance, value):
+        await self.motor_done_move.write(value='False')
+        await self.user_readback.write(value=value)
```
tacaswell: This should probably use something like https://github.com/NSLS-II/caproto/pull/467/files#diff-0e7b649e7b9201b19f93425c52a4f370R58 to simulate motion.

nslsii/iocs/tests/temperature_controllers_test.py
```diff
@@ -26,14 +27,22 @@ def test_Eurotherm(RE):
...
 
     # Start up an IOC based on the thermo_sim device in caproto.ioc_examples
+    '''
     ioc_process = subprocess.Popen([sys.executable, '-m',
                                     'caproto.tests.example_runner',
                                     'caproto.ioc_examples.thermo_sim'],
                                    stdout=stdout, stdin=stdin,
                                    env=os.environ)
+    '''
+    ioc_process = subprocess.Popen([sys.executable, '-m',
```
awalter-bnl: Can you provide an explanation for why this change is necessary and remove the commented out section above?

nslsii/iocs/tests/test_epicsmotor_ioc.py
```diff
@@ -0,0 +1,102 @@
...
+    stdout = subprocess.PIPE
+    stdin = None
+
+    '''
+    ioc_process = subprocess.Popen([sys.executable, '-m',
+                                    'caproto.tests.example_runner',
+                                    'nslsii.iocs.epics_motor_ioc_sim'],
+                                   stdout=stdout, stdin=stdin,
+                                   env=os.environ)
+    '''
```
awalter-bnl: see comment above, why is this change necessary and can we remove the commented out code above

nslsii/iocs/epics_motor_ioc_sim.py
```diff
@@ -0,0 +1,161 @@
...
+                              enum_strings=_cmd_states,
+                              dtype=ChannelType.ENUM,
+                              name='.HOMR')
+
+    # Methods
+
+    @user_setpoint.putter
+    async def user_setpoint(self, instance, value):
+        p0 = instance.value
+        dwell = self._step_size/self._velocity
```
awalter-bnl: Nice, I like that the time it takes to perform the scan is related to the velocity.

nslsii/iocs/tests/test_epstwostate_ioc.py
```diff
@@ -43,11 +43,17 @@ def test_epstwostate_ioc():
     stdout = subprocess.PIPE
     stdin = None
 
+    '''
     ioc_process = subprocess.Popen([sys.executable, '-m',
                                     'caproto.tests.example_runner',
                                     'nslsii.iocs.eps_two_state_ioc_sim'],
                                    stdout=stdout, stdin=stdin,
                                    env=os.environ)
+    '''
```
awalter-bnl: As above please remove commented out sections and add a description of why this change is necessary

nslsii/iocs/epics_motor_ioc_sim.py
malitsky: user_setpoint() was updated with a similar scenario

nslsii/iocs/tests/temperature_controllers_test.py
malitsky: it removed an additional indirection. Today, temperature_contollers_test however failed in both cases (with and without caproto.tests.example_runner) even when run as the first test. Therefore, in this PR it was disabled by renaming the corresponding file.

nslsii/iocs/tests/test_epstwostate_ioc.py
malitsky: See above comments associated with temperature_contollers_test

nslsii/iocs/tests/test_epicsmotor_ioc.py
malitsky: See above comments associated with temperature_contollers_test

nslsii/iocs/tests/temperature_controllers_test.py
awalter-bnl: OK so I don't follow this, and what I take form this line:
> Therefore, in this PR it was disabled by renaming the corresponding file.

is that we are now no longer testing against that test which we will need to resolve before merging.

I am not comfortable with renaming the test and placing it in `nslsii/iocs/tests` as this is not a test of the ioc but a test of the ophyd class, which logically belongs in `nslsii/tests/`.

Does it help if we move all of the tests (even the ioc tests) to `nslsii/tests`? It is our standard way of arranging the repos, with all tests in `nslsii/tests` and then seperated into modules (say a single file for all `ioc tests`).

awalter-bnl (changes requested):
I would like to resolve the confusion and issues with the tests prior to merging see above comment chain.

nslsii/iocs/tests/temperature_controllers_test.py
malitsky: all files have been moved to nslsii/tests and successfully tested

nslsii/iocs/epics_motor_ioc_sim.py
```diff
@@ -0,0 +1,161 @@
...
+        N = max(1, int((value - p0) / self._step_size))
+
+        await self.motor_done_move.write(value='False')
+
+        for j in range(N):
+            new_value = p0 + self._step_size*(j+1)
+            time.sleep(dwell)
```
tacaswell: Please do not use sychronous sleeps in async functions, it will block the server while it is dwelling.

nslsii/iocs/epics_motor_ioc_sim.py
```diff
@@ -0,0 +1,161 @@
...
+                              name='.HOMF')
+    home_reverse = pvproperty(value=_cmd_states[0],
+                              enum_strings=_cmd_states,
+                              dtype=ChannelType.ENUM,
+                              name='.HOMR')
+
+    # Methods
+
+    @user_setpoint.putter
+    async def user_setpoint(self, instance, value):
```
tacaswell: what happens if this is set twice with very different values, say it starts at 0 and I set -100 and 100 ~simultaneously from different clients?

nslsii/tests/temperature_controllers_test.py
```diff
@@ -26,14 +27,22 @@ def test_Eurotherm(RE):
     stdin = None
 
     # Start up an IOC based on the thermo_sim device in caproto.ioc_examples
+    '''
```
tacaswell: please do not comment things out by putting strings around them, it makes the diff very hard to read.  Either remove the old code (it is under version control so we can get it back!) or put `#` at the beginning of each line (most editors have helpers to do this in bulk).

nslsii/tests/test_epicsmotor_ioc.py
```diff
@@ -0,0 +1,95 @@
...
+        # 3. move (timeout > moving time)
+
+        target_val = 7
+        mvtime = (target_val - readback_val)/velocity_val
+
+        move_status = MoveStatus(mtr, target_val)
+
+        try:
+            move_status = mtr.move(target_val, timeout=mvtime+1)
+        except RuntimeError:
```
tacaswell: why are we allowing `RuntimeError`s to pass?  If we expect them please use `pytest.raises`, if we do not shouldn't that make the test fail?

nslsii/tests/test_epicsmotor_ioc.py
```diff
@@ -0,0 +1,95 @@
...
+        move_status = MoveStatus(mtr, target_val)
+
+        try:
+            move_status = mtr.move(target_val, timeout=mvtime-1)
+        except RuntimeError:
+            pass
+
+        assert move_status.success is False
+
+    finally:
```
tacaswell: Please do this clean up with a fixture.

nslsii/tests/test_epicsmotor_ioc.py
```diff
@@ -0,0 +1,95 @@
...
+    time.sleep(5)
+
+    # Wrap the rest in a try-except to ensure the ioc is killed before exiting
+    try:
+
+        mtr = EpicsMotor(prefix='mtr:', name='mtr')
+
+        time.sleep(5)
+
+        # 1. check the ioc-device connection and initial values
```
tacaswell: it looks like each of these should be its own test?

awalter-bnl (dismissed):

awalter-bnl (commented):
@klauer thanks for the review request, however I already have reviewed this a few times. The last concerns I had related to the location of the test files which @malitsky resolved, hence why I approved it. 

Clearly their are still some outstanding comments from @tacaswell that require resolution.

tacaswell (changes requested):
The test should be split up into multiple tests and the clean up logic moved to a fixture.

The `time.sleep` needs be removed.

The questions about how this behaves to semi-pathological input should be answered, but we can discuss if fixing that is in scope for this or not.

nslsii/iocs/epics_motor_ioc_sim.py
malitsky: added putter_lock

malitsky: added putter_lock

malitsky:
PR updated with Tom's comments.

malitsky:
fyi, @tacaswell and @awalter-bnl

PR #67, opened 2019-05-23
LakeshoreIOC

malitsky:


awalter-bnl (approved):
This looks good to me, nice work implementing the damped-sin path for the readback temperatures.

nslsii/iocs/lakeshore_control.py
```diff
@@ -0,0 +1,240 @@
...
+    async def done(self, instance):
+        return self._done_val
+
+    @target_channel.getter
+    async def target_channel(self, instance):
+        return self._target_channel_val
+
+    @target_channel.putter
+    async def target_channel(self, instance, value):
+        self._target_channel_val = value
```
klauer: Additional attributes for these should be unnecessary; you could replace:
`self._target_channel_val` with `self.target_channel.value` (for reading) or `await self.target_channel.write(value=0)` for writing.

malitsky: The extra layer was added to implement potential extensions for simulating hardware issues.

PR #68, opened 2019-06-05
DOC: release

ke-zhang-rd:
Create release-history file and document v0.0.10

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -1,0 +1,10 @@
+***************
+Release History
+***************
+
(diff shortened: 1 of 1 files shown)


docs/source/release-history.rst
```diff
@@ -0,0 +1,10 @@
+***************
+Release History
+***************
+
+v0.0.10 (2019-06-04)
```
danielballan: Date needs to be updated once you are ready to actually tag. Otherwise looks good!

PR #69, opened 2019-06-06
BLD: pin attrs>=17.4.0

ke-zhang-rd:
For error report below
https://travis-ci.com/NSLS-II/nslsii/jobs/205765574

Solution are based on discussion here 
https://github.com/Julian/jsonschema/issues/449


danielballan:
Wouldn't the fix in databroker have propagated here and been sufficient? Unlike `databroker`, `nslsii` has no direct dependency on `attrs`.

PR #70, opened 2019-08-06
Add missing files from sdist

mrakitin:
We are missing a couple of files in the source distribution. This was noticed when a sdist tar.gz package was pushed to PyPI without the corresponding wheel package: https://pypi.org/project/nslsii/0.0.10/#files.

The `setup.py` was updated based on databroker's [setup.py](https://github.com/bluesky/databroker/blob/cebed19d83445fb2609b33cf8ec380eadc50f171/setup.py#L4-L14). with slight modifications.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,4 @@
+include README.md
+include requirements.txt
 include versioneer.py
 include nslsii/_version.py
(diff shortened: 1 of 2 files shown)


mrakitin:
Thanks!

PR #71, opened 2019-08-09
Update pv suffixes

awalter-bnl:
The IOC for the Eurotherm was updated some time ago but the ophyd object for it was not updated to mirror the new PV suffixes.

### DIFF
--- a/nslsii/iocs/thermo_sim.py
+++ b/nslsii/iocs/thermo_sim.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python3
+from caproto.server import pvproperty, PVGroup, ioc_arg_parser, run
+import numpy as np
+import time
+from textwrap import dedent
+
+
+class Thermo(PVGroup):
+    """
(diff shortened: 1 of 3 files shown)


mrakitin:
Looks good, but the tests failed:
```
self = <[TimeoutError("Failed to connect to thermo:T-SP") raised in repr()] EpicsSignal object at 0x7f09471c0748>
```
Is it something to be fixed in the tests as well?

awalter-bnl:
Ah good catch @mrakitin. I will have to put in a PR to caproto to update the simulated Eurotherm IOC with the new PR’s I will do that on Monday

awalter-bnl:
A PR to update the caproto simulated IOC has been placed into `caproto/caproto` PR#501 [here](https://github.com/caproto/caproto/pull/501)

This should resolve the test failures we are observing by updating to the new PV suffixes.

awalter-bnl:
@mrakitin I now have the tests passing. can you take another look?

PR #72, opened 2019-08-21
ENH: Baton IOC

tacaswell:
This is the companion PR to
https://github.com/bluesky/bluesky/pull/1246

Adds:

 - updates configure_base for bluesky changes
 - a Baton caproto IOC and matching ohpyd object with the hooks to
   interact with bluesky
 - a read only version of the ophyd object to back a typhon window

```python

from nslsii import configure_base
from nslsii.baton import Baton
from ophyd.sim import *


b = Baton('XF31ID:', name='b')

names = configure_base(get_ipython().user_ns, 'temp', baton=b)


```

```python
import os
os.environ['PYDM_EPICS_LIB'] = 'caproto'
from nslsii.baton import BatonDisplay

from qtpy.QtWidgets import QApplication
import typhon
app = QApplication.instance() or QApplication(['bluesky'])

bd = BatonDisplay('XF31ID:', name='b')
suite = typhon.TyphonSuite.from_device(bd)
suite.show()

app.exec_()

```

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -118,7 +118,7 @@ def configure_base(user_ns, broker_name, *,
         RE = RunEngine(md, **kwargs)
 
         if baton is not None:
-            for d in ['start', 'stop']:
+            for d in ['start']:
                 tok = RE.subscribe(baton.doc_callback, d)
                 baton.tokens.append(tok)
             RE.state_hook = baton.state_callback
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -13,7 +13,7 @@ def import_star(module, ns):
 
 def configure_base(user_ns, broker_name, *,
                    bec=True, epics_context=False, magics=True, mpl=True,
-                   ophyd_logging=True, pbar=True):
+                   ophyd_logging=True, pbar=True, baton=None):
     """
     Perform base setup and instantiation of important objects.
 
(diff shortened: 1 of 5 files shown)


prjemian:
Somewhere, can you define in very plain terms, what is a `baton`?

tacaswell:
I am a former track runner and orchestra player so I am either thinking the relay baton or a conductors baton.

It's the thing that means you are in charge :wink:

prjemian:
Got it.  So in the context of how Bluesky is used, the process in posession of the figurative *baton* is the only one authorized to control the hardware on the instrument (beam line, whatever).  Any other bluesky process must be passive and not command the hardware to change.  Right?

prjemian:
Such as this example: https://github.com/APS-2BM-MIC/ipython-user2bmb/blob/master/profile_2bmb/startup/20-signals.py#L24-L37

prjemian:
Later in that file is the use of the *baton*: https://github.com/APS-2BM-MIC/ipython-user2bmb/blob/master/profile_2bmb/startup/20-signals.py#L52-L95

Basically, if *has baton*, attach to hardware, else attach to simulators.

tacaswell:
@prjemian Yes, same idea.  That looks like that is an end-station / instrument deconfliction tool, this is to prevent a second RE on the same instrument from starting up and if it get stolen (which being EPICS is something we can not actually protect against), prevents an existing RE from running.

Could you push https://github.com/APS-2BM-MIC/ipython-user2bmb/blob/9b7f7b134d0173cfd3a3398b79146cf7478071d2/profile_2bmb/startup/18-suspenders.py#L8-L58 upstream to bluesky?  That looks generically useful.

prjemian:
Sure

PR #73, opened 2019-08-29
IPython exception logger

jklynch:
This PR adds one module with one function for logging exceptions to the IPython log file.

The new module is nslsii.common.ipynb.logutils. Should it be in nslsii.common instead?

Also, is there a way to write tests against IPython?

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -13,7 +13,7 @@ def import_star(module, ns):
 
 def configure_base(user_ns, broker_name, *,
                    bec=True, epics_context=False, magics=True, mpl=True,
-                   ophyd_logging=True, pbar=True):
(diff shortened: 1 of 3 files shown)


tacaswell:
Could you also add a call to this to `base_config`?

I added some ipython driven tests in https://github.com/bluesky/bluesky/pull/1242/files.

You can do something like

```python
from mock import MagicMock
ip = IPython.core.interactiveshell.InteractiveShell()
ip.logger = MagicMock()
ip.set_custom_exc((BaseException, ), log_exception)
￼ip.run_cell("raise Exception")
assert ip.logger.log_wirte.call_count > 0
```

jklynch:
added a test

mrakitin:
It's good to have a test. We can use it as an example in the future.

PR #77, opened 2020-01-21
Protect against double subscription

### LINKED ISSUE #76: Avoid double subscription of databroker to RE
Motivated by:
- https://github.com/NSLS-II/Bug-Reports/issues/203
- https://controlsweb.nsls2.bnl.gov/trac/ticket/4323

The problem happened due to the second manual explicitly called `%run -i 00-startup.py`.

danielballan:
Closes #76

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -79,6 +79,10 @@ def configure_base(user_ns, broker_name, *,
 
     >>>> configure_base(get_ipython().user_ns, 'chx');
     """
+    if configure_base.has_been_called:
+        raise RuntimeError(
+            "configure_base should only be called once per process.")
+    configure_base.has_been_called = True
     ns = {}  # We will update user_ns with this at the end.
 
     # Set up a RunEngine and use metadata backed by a sqlite file.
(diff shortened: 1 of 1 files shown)


tacaswell:
I think it would be better to stash this in the namespace passed into `configure_base`, not in the nslsii module.

danielballan:
True. That approach is better because it allows for the possibility of running multiple times as long as the namespaces are separate (good for tests) and it works even if the user does `reload(nslsii)`.

nslsii/__init__.py
```diff
@@ -80,7 +80,12 @@ def configure_base(user_ns, broker_name, *,
     >>>> configure_base(get_ipython().user_ns, 'chx');
     """
     ns = {}  # We will update user_ns with this at the end.
-
+    # Protect against double-subscription.
+    SENTINEL = "__nslsii_configure_base_has_been_run"
+    if user_ns.get(SENTINEL)
```
klauer:
```suggestion
    if user_ns.get(SENTINEL):
```

PR #78, opened 2020-02-18
Add transforms

danielballan:
This is a work in progress, will be expanded as we find more transforms that we need.

### DIFF
--- a/nslsii/transforms.py
+++ b/nslsii/transforms.py
@@ -3,7 +3,7 @@
 import copy
 
 
-def fix_csx_scaler_shape(d):
+def csx_fix_scaler_shape(d):
     """
     Transform an Event Descriptor.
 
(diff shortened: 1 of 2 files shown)


nslsii/transforms.py
```diff
@@ -0,0 +1,11 @@
+# DataBroker "transforms" for patching up documents
+# on their way out.
+import copy
+
+
+def fix_csx_scaler_shape(d)
```
gwbischof: missing a :

gwbischof:
What if we put the beamline name first? And no beamline name if it can apply to other beamlines?

nslsii/transforms.py
danielballan: D'oh!

danielballan:
That works for me. Change pushed.

PR #79, opened 2020-03-03
Sync xspress3 code with hxntools

mrakitin:
This update is to have the code for xspress3 synched with hxntools.

An extra update is to correct the import of `DeviceStatus` to work properly with ophyd 1.4.x.


mrakitin:
@tacaswell, bump.

tacaswell:
I'm not sure I can reason through this without having a device to play with.

mrakitin:
Thanks @tacaswell! I tested this code on the ISS' xspress3 detector, and it worked fine. See https://github.com/NSLS-II-ISS/profile_collection/pull/40/files#diff-b86b71b8ada3f1c237f9c32061df2d92 for details.

PR #81, opened 2020-03-04
Add convenience func for migrating RE.md.

danielballan:


danielballan:
This is follow-up to #75, turns the work of copying metadata from the old system to the new system into:

```py
from nslsii import migrate_metadata
migrate_metadata()
```

cc @mrakitin who recently did this manually at SIX

Issue #82, opened 2020-03-13
Add a transform for SST


danielballan:
Base it on https://github.com/bluesky/databroker/pull/474

PR #83, opened 2020-03-17
added transform to fix rsoxs descriptor

gwbischof:
I tested a bit with real rsoxs data.. until i ran out of RAM.

### DIFF
--- a/nslsii/transforms.py
+++ b/nslsii/transforms.py
@@ -3,6 +3,21 @@
 import copy
 
 
+def remove_zeros(shape):
+    """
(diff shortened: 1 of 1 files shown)


gwbischof:
Fixes #82

gwbischof:
based on the PR: https://github.com/bluesky/databroker/pull/474 I thought I was going to have to swap the x,y dims. But i was getting shape errors when I did that.

gwbischof:
I made the helper functions, because I suspect that other descriptor transforms will use them too

PR #86, opened 2020-06-01
Add function to subscribe a Kafka Publisher to the RunEngine

jklynch:
This PR adds functions for working with Kafka.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -208,7 +208,7 @@ def configure_base(
         subscribe_kafka_publisher(
             RE,
             beamline_name=broker_name,
-            bootstrap_servers="10.0.137.8:9092,10.0.137.21:9092,10.0.137.22:9092",
+            bootstrap_servers="cmb01:9092,cmb02:9092,cmb03:9092",
             producer_config={
                 "enable.idempotence": True,
                 "linger.ms": 0
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -520,5 +520,6 @@ def subscribe_kafka_publisher(RE, beamline_name, bootstrap_servers):
          }
     )
     subscription_token = RE.subscribe(_kafka_publisher)
-    print(f'RE will publish documents to Kafka topic {topic}')
+    logging.getLogger("nslsii").info('RE will publish documents to Kafka topic %s', topic)
+
     return subscription_token
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -208,7 +208,7 @@ def configure_base(
         subscribe_kafka_publisher(
             RE,
             beamline_name=broker_name,
-            bootstrap_servers="10.0.137.8:9092"
+            bootstrap_servers="10.0.137.8:9092,10.0.137.21:9092,10.0.137.22:9092"
         )
 
     # always configure %xmode minimal
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -477,10 +477,7 @@ def migrate_metadata():
     new_md.update(old_md)
 
 
-_kafka_publisher = None
-
-
-def subscribe_kafka_publisher(RE, beamline_name):
+def subscribe_kafka_publisher(RE, beamline_name, bootstrap_servers):
     """
     Create and subscribe a Kafka Publisher to a RunEngine. Keep a reference to the Publisher.
     The Publisher will publish to Kafka topic "<beamline>.bluesky.documents".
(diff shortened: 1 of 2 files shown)


nslsii/__init__.py
```diff
@@ -471,3 +475,48 @@ def migrate_metadata():
...
+    Parameters
+    ----------
+    RE: RunEngine
+
+    """
+    global _kafka_publisher
+    if _kafka_publisher is None:
+        print("No Kafka Publisher exists")
+    else:
+        RE.unsubscribe(_kafka_publisher)
```
danielballan, line 522: I do not think this will work the way you intend. For rather obscure reasons, the `unsubscribe` method expects the `token` (integer) returned by `RE.subscribe(...)`, not the callable itself. To avoid global state, it might be reasonable to do

```py
def subscribe_kafka_publisher(RE, beamline_name):
    ...
    token = RE.subscribe(...)
    return token
```

and leave for downstream code to receive the `token` and manage subscription in the normal way --- `RE.unsubscribe(token)`.

gwbischof:
I think that _kafka_publisher should be on a thread so that it doesn't block, and in a try-except so that is doesn't interrupt an experiment if it fails.

gwbischof:
Here is an example: https://github.com/NSLS-II/nslsii/blob/176a45802a9413660f9e1f424d7ae8bed227497c/nslsii/__init__.py#L454

gwbischof:
What about making a decorator that runs the function on a thread with try-except?

danielballan:
Extending @gwbischof's suggestion --- utility decorators to do this (for a simple function callback and class-based `DocumentRouter` callback, respectively) were recently added to bluesky by @tacaswell but I think not yet documented. They are:
https://github.com/bluesky/bluesky/blob/4fab894bddbd4a563f28852ea3171b87140ae7b9/bluesky/callbacks/core.py#L21-L117

But these decorators just catch and log exceptions; they aren't robust against callbacks that might hang for a long time or get very far behind. A variation could be added here (and upstream-ed later) that gives the extra protection of a queue + background thread, as implemented for olog.

As an aside, the need for this fault-tolerant subscription is probably a temporary state of affairs because in the future we can just rely on Kafka for fault tolerance. Specifically:
* The Kafka subscriber will become essential: if it fails, we want to raise and error because data won't be being saved!
* The olog subscriber may as well move downstream to consume from Kafka, to remove load from the main RunEngine process. Thus, if it gets behind or fails for awhile, Kafka will handle that itself.

nslsii/__init__.py
```diff
@@ -471,3 +475,40 @@ def migrate_metadata():
...
+    _kafka_publisher = Publisher(
+         topic=topic,
+         bootstrap_servers=bootstrap_servers,
+         key=uuid.uuid4(),
+         producer_config={
+             "enable.idempotence": True
+         }
+    )
+    subscription_token = RE.subscribe(_kafka_publisher)
+    print(f'RE will publish documents to Kafka topic {topic}')
```
danielballan, line 513: Questionable whether we want this to print at `IPython` startup.

nslsii/__init__.py
```diff
@@ -471,3 +475,40 @@ def migrate_metadata():
     os.makedirs(directory, exist_ok=True)
     new_md = PersistentDict(directory)
     new_md.update(old_md)
+
+
+def subscribe_kafka_publisher(RE, beamline_name, bootstrap_servers):
```
danielballan, line 480: Should we hard-code a default set of `bootstrap_servers`? If we don't, we'll have to hard-code them in every beamline profile, which will make them harder to change.

nslsii/__init__.py
jklynch: I will log it instead.

nslsii/__init__.py
jklynch: I think this would be a reasonable place to list the bootstrap servers. There is still a question in my mind about running a kafka broker at each beamline to cache messages in the event connectivity to the kafka cluster is lost.
Either way we could list the CMB machines right now, but does that make sense?

jklynch: I added CMB01, CMB02, and CMB03 to the bootstrap server list.

jklynch:
I am doing something kind of questionable in using `broker_name` to get the beamline name. It would be nice to have a better way to get the beamline name.

nslsii/__init__.py
```diff
@@ -197,6 +204,13 @@ def configure_base(
             exception_logger=log_exception, ipython=get_ipython()
         )
 
+    if publish_documents_to_kafka:
+        subscribe_kafka_publisher(
+            RE,
+            beamline_name=broker_name,
+            bootstrap_servers="10.0.137.8:9092,10.0.137.21:9092,10.0.137.22:9092"
```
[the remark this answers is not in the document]
jklynch: That sounds reasonable.

nslsii/__init__.py
```diff
@@ -197,6 +204,17 @@ def configure_base(
...
         )
 
+    if publish_documents_to_kafka:
+        subscribe_kafka_publisher(
+            RE,
+            beamline_name=broker_name,
+            bootstrap_servers="cmb01:9092,cmb02:9092,cmb03:9092",
+            producer_config={
+                "enable.idempotence": True,
+                "linger.ms": 0
```
danielballan, line 214: This is the default, right? https://docs.confluent.io/current/installation/configuration/producer-configs.html#linger.ms

PR #87, opened 2020-06-03
ENH: Add prompt.py

ke-zhang-rd:


mrakitin:
I guess the example was taken from https://github.com/cookiecutter/cookiecutter/blob/master/cookiecutter/prompt.py. Just for reference and to give credits to the devs.

ke-zhang-rd:
> I guess the example was taken from https://github.com/cookiecutter/cookiecutter/blob/master/cookiecutter/prompt.py. Just for reference and to give credits to the devs.

Thanks.

Issue #84, opened 2020-03-23
provide a reasonable bluesky log file path for Windows


jklynch:
The bluesky log path is /var/log/bluesky/bluesky.log. This is not reasonable for Windows.

danielballan:
The widely-used solution for this is https://pypi.org/project/appdirs/

For the sake of prioritization, do we have any Windows bluesky deployments left now that SST moved to Linux? I think there might be one, but I can't immediately think of it.

tacaswell:
This library is endorsed by at least one windows based core cpython dev.

PR #88, opened 2020-06-03
use appdirs to determine default logging directory

jklynch:
The default logging directory will be determined by appdirs.user_log_dir(). Environment variables BLUESKY_LOG_FILE and BLUESKY_IPYTHON_LOG_FILE can still be used to override the default.

This PR addresses issue #84.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -35,7 +35,7 @@ def configure_base(
     mpl=True,
     configure_logging=True,
     pbar=True,
-    ipython_exc_logging=True,
+    ipython_logging=True,
 ):
     """
     Perform base setup and instantiation of important objects.
@@ -80,13 +80,12 @@ def configure_base(
         True by default. Set False to skip matplotlib ``ion()`` at event-loop
         bridging.
     configure_logging : boolean, optional
-        True by default. Set False to skip INFO-level logging to
-        /var/logs/bluesky/bluesky.log.
+        True by default. Set False to skip INFO-level logging.
     pbar : boolean, optional
         True by default. Set false to skip ProgressBarManager.
-    ipython_exc_logging : boolean, optional
-        True by default. Exception stack traces will be written to IPython
-        log file when IPython logging is enabled.
+    ipython_logging : boolean, optional
+        True by default. Console output and exception stack traces will be
+        written to IPython log file when IPython logging is enabled.
 
     Returns
     -------
@@ -193,7 +192,7 @@ def configure_base(
     if configure_logging:
         configure_bluesky_logging(ipython=get_ipython())
 
-    if ipython_exc_logging:
+    if ipython_logging:
         from nslsii.common.ipynb.logutils import log_exception
 
         # IPython logging will be enabled with logstart(...)
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -377,7 +377,12 @@ def configure_ipython_logging(
         bluesky_ipython_log_file_path.rename(
             str(bluesky_ipython_log_file_path) + ".old"
         )
-    ipython.magic(f"logstart -o -t {bluesky_ipython_log_file_path} append")
+    # ipython gives a warning if logging fails to start, for example if the log
+    # directory does not exist. Convert that warning to an exception here.
+    with warnings.catch_warnings():
+        warnings.simplefilter(action="error")
+        # specify the file for ipython logging output
+        ipython.magic(f"logstart -o -t {bluesky_ipython_log_file_path} append")
 
     return bluesky_ipython_log_file_path
 
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -2,7 +2,10 @@ from distutils.version import LooseVersion
 import logging
 from logging.handlers import TimedRotatingFileHandler
 import os
+from pathlib import Path
 import sys
+
+import appdirs
 
 from IPython import get_ipython
 from ._version import get_versions
@@ -244,7 +247,7 @@ def configure_base(
     return list(ns)
 
 
-def configure_bluesky_logging(ipython):
+def configure_bluesky_logging(ipython, appdirs_appname="bluesky"):
     """
     Configure a TimedRotatingFileHandler log handler and attach it to
     bluesky, ophyd, nslsii, and ipython loggers.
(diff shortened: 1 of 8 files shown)


jklynch:
Why is test_Eurotherm failing?

nslsii/__init__.py
```diff
@@ -256,30 +259,37 @@ def configure_bluesky_logging(ipython):
...
     -------
-    bluesky_log_file_path: str
+    bluesky_log_file_path: Path
         log file path
 
     """
     global bluesky_log_file_path
 
     if "BLUESKY_LOG_FILE" in os.environ:
+        bluesky_log_file_path = Path(os.environ["BLUESKY_LOG_FILE"])
```
mrakitin, line 275: Should we validate somehow that the path passed via the environment variable is correct?

nslsii/__init__.py
```diff
@@ -256,30 +259,37 @@ def configure_bluesky_logging(ipython):
...
             file=sys.stderr,
         )
-        bluesky_log_file_path = os.environ["BLUESKY_LOG_FILE"]
     else:
+        bluesky_log_file_path = Path(
+            appdirs.user_log_dir(appname=appdirs_appname)
+        ) / Path("bluesky.log")
         print(
-            f"environment variable BLUESKY_LOG_FILE is not set, using default file path",
+            "environment variable BLUESKY_LOG_FILE is not set, using default file path",
```
mrakitin, line 285: I'd print the default file path here, just to be explicit.

nslsii/__init__.py
```diff
@@ -320,38 +332,45 @@ def configure_ipython_exc_logging(exception_logger, ipython, rotate_file_size=10
...
     -------
-    bluesky_ipython_log_file_path: str
+    bluesky_ipython_log_file_path: Path
         log file path
 
     """
     ipython.set_custom_exc((BaseException,), exception_logger)
 
     if "BLUESKY_IPYTHON_LOG_FILE" in os.environ:
+        bluesky_ipython_log_file_path = Path(os.environ["BLUESKY_IPYTHON_LOG_FILE"])
```
mrakitin, line 348: Same question as above for existence checking.

nslsii/__init__.py
```diff
@@ -320,38 +332,45 @@ def configure_ipython_exc_logging(exception_logger, ipython, rotate_file_size=10
...
     # if a previous copy exists just overwrite it
     if (
-        os.path.exists(bluesky_ipython_log_file_path)
+        bluesky_ipython_log_file_path.exists()
         and os.path.getsize(bluesky_ipython_log_file_path) >= rotate_file_size
     ):
-        os.replace(
-            bluesky_ipython_log_file_path, bluesky_ipython_log_file_path + ".old"
+        bluesky_ipython_log_file_path.rename(
+            str(bluesky_ipython_log_file_path) + ".old"
```
mrakitin, line 373: Why not to use `TimedRotatingFileHandler` here?

nslsii/__init__.py
tacaswell: We were not checking before :shrug: .

I guess we could check that it exists (unless the logger will create it) and it is writabel?

mrakitin: Yes, I think it's best to check as least that the path is valid. It's a good point about the writable path.

nslsii/__init__.py
```diff
@@ -256,30 +259,37 @@ def configure_bluesky_logging(ipython):
...
+        bluesky_log_file_path = Path(
+            appdirs.user_log_dir(appname=appdirs_appname)
+        ) / Path("bluesky.log")
         print(
-            f"environment variable BLUESKY_LOG_FILE is not set, using default file path",
+            "environment variable BLUESKY_LOG_FILE is not set, using default file path",
             file=sys.stderr,
         )
-        bluesky_log_file_path = "/var/log/bluesky/bluesky.log"
+    if not bluesky_log_file_path.parent.exists():
```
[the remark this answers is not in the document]
tacaswell: For deployment we should be setting the envs, this is so the out-of-the-box usage works (for example on my system I don't have write permissions to `/var/log` as me).  I think the default behavior should work rather than be tuned to our deployment details.

nslsii/__init__.py
jklynch: I added tests for the case of a log directory that does not exist. The result is FileNotFoundError. I think this is the behavior we want rather than creating the directory, since the most probable cause is a mis-spelled log directory (someone knew what they wanted but a transmission error occurred at the keyboard).

I also added tests for the case of an un-writeable log directory. The result is PermissionError. Again, I think this is the behavior we want.

So while it seems like the proposed checks are the responsible thing to do, I don't think we can improve on the exceptions we already get.

jklynch: I told a lie :(. The ipython exception logger raises a warning rather than an error if the specified log directory does not exist.

I am converting that warning to an exception with `warnings.catch_warnings` and `warning.simplefilter("error")`.

nslsii/__init__.py
jklynch: I was printing it and then I noticed ipython also prints it, so I removed mine.

But I think you are right that this is also a good place to display the file path so I have added it.

nslsii/__init__.py
jklynch: see answer above

nslsii/__init__.py
jklynch: We can only submit an exception logging *function* to ipython. We can't mess with the logging internals (as far as I can tell).

nslsii/__init__.py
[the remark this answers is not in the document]
jklynch: We do point the ophyd and caproto loggers at the same file as the bluesky loggers. We are not doing that for databroker. Hmm. @danielballan should we point databroker loggers at the bluesky log file?

I would point LSDC logging at /var/log/LSDC, if it were up to me.

nslsii/__init__.py
jklynch: I should have mentioned the `bluesky_ipython_log_file_path` gets *all* ipython output, so it is not really a target for `logging`.

nslsii/__init__.py
mrakitin: Thanks.

nslsii/__init__.py
jklynch: Now that I think about it the name `configure_ipython_exc_logging` is *wrong*, since it configures both normal output and exception logging. I am changing it to `configure_ipython_logging` and adding some clarification to the docstring.

nslsii/__init__.py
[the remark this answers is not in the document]
danielballan: Surfacing a comment publicly:

Yes, but it's not urgent because databroker currently does little to no logging.

jklynch:
the only test failure is test_Eurotherm, still

danielballan:
Pushed a commit to mark the problematic Eurotherm test that we are consistently ignoring as a "known failure" so that CI red/green is meaningful again.

danielballan:
Because of the consistent Eurotherm failures we didn't notice flake8 errors. Added another commit fixing those....

danielballan:
Green!

PR #89, opened 2020-06-26
create default log directory

### LINKED ISSUE #84: provide a reasonable bluesky log file path for Windows

jklynch:
Closes #84.

This PR changes the behavior of the two logging configuration functions.
Previously the default log directory would not be created and an exception was raised if it did not exist.
Now if that directory does not exist it will be created.

Existing tests have been updated. New tests have been added.

### DIFF
--- a/nslsii/tests/test_logutils.py
+++ b/nslsii/tests/test_logutils.py
@@ -57,7 +57,39 @@ def test_configure_bluesky_logging_with_unwriteable_dir(tmpdir):
         configure_bluesky_logging(ipython=ip,)
 
 
-def test_default_bluesky_logging():
+def test_configure_bluesky_logging_creates_default_dir():
+    """
+    Remove environment variable BLUESKY_LOG_FILE and test that
+    the default log file path is created. This test creates a
+    directory rather than using pytest's tmp_path so the test
+    must clean up at the end.
+    """
+    test_appname = "bluesky-test"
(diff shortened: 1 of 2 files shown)


nslsii/tests/test_logutils.py
```diff
@@ -57,24 +57,51 @@ def test_configure_bluesky_logging_with_unwriteable_dir(tmpdir):
...
     test_appname = "bluesky-test"
-    user_log_dir = Path(appdirs.user_log_dir(appname=test_appname))
+    log_dir = Path(appdirs.user_log_dir(appname=test_appname))
+    # remove log_dir if it exists
+    # make a reasonable effort to delete the log directory
+    # no heroics
+    if log_dir.exists():
+        for f in log_dir.iterdir():
+            f.unlink()
+        log_dir.rmdir()
```
mrakitin, line 75: Why not to use https://docs.python.org/3/library/shutil.html#shutil.rmtree here? Seems simpler.

MikeHart85, line 75: Is there any danger of inadvertently deleting logs that would have been useful to keep with this? Or is this appname only used here?

jklynch: The appname here is intended to be testing-specific.

PR #91, opened 2020-07-01
ENH: make minimal traceback reporting optional

tacaswell:
Because sometimes you do want to see the full tracebacks.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -44,6 +44,7 @@ def configure_base(
     pbar=True,
     ipython_logging=True,
     publish_documents_to_kafka=False,
+    tb_minimize=True,
 ):
     """
     Perform base setup and instantiation of important objects.
@@ -95,7 +96,9 @@ def configure_base(
         True by default. Console output and exception stack traces will be
         written to IPython log file when IPython logging is enabled.
     publish_documents_to_kafka: boolean, optional
-        False by default. If True publish bluesky documents to a Kafka message broker.
+        False by default. If True publish bluesky documents to a Kafka message broker
+    tb_minimize : boolean, optional
+        If IPython should print out 'minimal' tracebacks.
 
     Returns
     -------
@@ -221,9 +224,10 @@ def configure_base(
             },
         )
 
-    # always configure %xmode minimal
-    # so short tracebacks are printed to the console
-    get_ipython().magic("xmode minimal")
+    if tb_minimize:
+        # configure %xmode minimal
+        # so short tracebacks are printed to the console
+        get_ipython().magic("xmode minimal")
 
     # convenience imports
     # some of the * imports are for 'back-compatibility' of a sort -- we have


nslsii/__init__.py
```diff
@@ -92,7 +93,9 @@ def configure_base(
         True by default. Console output and exception stack traces will be
         written to IPython log file when IPython logging is enabled.
     publish_documents_to_kafka: boolean, optional
-        False by default. If True publish bluesky documents to a Kafka message broker.
+        False by default. If True publish bluesky documents to a Kafka message broker
+    tb_minimize : boolean, optional
```
mrakitin, line 97:
```suggestion
    tb_minimize: boolean, optional
```

nslsii/__init__.py
```diff
@@ -212,15 +215,14 @@ def configure_base(
...
-                "enable.idempotence": True,
-                "linger.ms": 0
-            }
+            producer_config={"enable.idempotence": True, "linger.ms": 0},
         )
 
-    # always configure %xmode minimal
-    # so short tracebacks are printed to the console
-    get_ipython().magic("xmode minimal")
+    # always
```
mrakitin, line 221: Was it left after moving comments around?

klauer:
It'd be nice if there were an exception stashing mechanism that worked like:

```
In [6]: %xmode minimal
Exception reporting mode: Minimal

In [7]: def test():
   ...:     raise Exception('oops')
   ...:

In [8]: test()
Exception: oops
(To see the full exception, run `%xstashed 4`)

In [9]: %xstashed 4
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-11-fbd55f77ab7c> in <module>
----> 1 test()
        global test = <function test at 0x10fa3a9e0>

<ipython-input-7-5f7df1f0ec85> in test()
      1 def test():
----> 2     raise Exception('oops')
        global Exception = undefined
      3

Exception: oops
```

Of course `%debug` could be used so long as you're only interested in the last raised exception.
At least with the above, even old exceptions could be reviewed interactively.
`</hijack>`

nslsii/__init__.py
tacaswell: numpydoc says there should be a space before the `:`.

nslsii/__init__.py
tacaswell: oops.

tacaswell:
@klauer That request should probably go to IPython

danielballan:
I think it exists: https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-tb

danielballan:
Ah, not exactly. I somewhat misinterpreted @klauer’s comment on my initial reading. Nonetheless that might be interesting.

klauer:
Just throwing it out there 🤷‍♂️

nslsii/__init__.py
mrakitin: Hm, it's conflicting with what we discovered recently with @danielballan. Dan, do you remember where we saw issues with the latest Sphinx?

mrakitin (approved):
Needed for the queueserver work.

PR #92, opened 2020-07-07
add function to subscribe KafkaPublishers to the RunEngine

jklynch:
This PR adds a function to subscribe Kafka Publishers to the RunEngine using a RunRouter.

A test for this function is also added which requires a Kafka broker to be running. The broker for testing is set up in the same way as for bluesky-kafka.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -43,7 +43,7 @@ def configure_base(
     configure_logging=True,
     pbar=True,
     ipython_logging=True,
-    publish_documents_to_kafka=False
+    publish_documents_to_kafka=False,
 ):
     """
     Perform base setup and instantiation of important objects.
@@ -206,19 +206,14 @@ def configure_base(
         from nslsii.common.ipynb.logutils import log_exception
 
         # IPython logging will be enabled with logstart(...)
-        configure_ipython_logging(
-            exception_logger=log_exception, ipython=get_ipython()
-        )
+        configure_ipython_logging(exception_logger=log_exception, ipython=get_ipython())
 
     if publish_documents_to_kafka:
         subscribe_kafka_publisher(
             RE,
             beamline_name=broker_name,
             bootstrap_servers="cmb01:9092,cmb02:9092,cmb03:9092",
-            producer_config={
-                "enable.idempotence": True,
-                "linger.ms": 0
-            }
+            producer_config={"enable.idempotence": True},
         )
 
     # always configure %xmode minimal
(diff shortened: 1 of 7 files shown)


jklynch:
I finally found a configuration parameter that fixed a problem: `auto.commit.interval.ms` is relevant to CI testing and may turn out to be important elsewhere. I was having an issue with this Kafka test getting double the expected number of documents from Kafka. The extra documents were left over from the previous test run (so the very first time I ran the test there was no issue) because the consumer did not signal to the broker it had processed them, even though it had. By default consumers can wait up to 5 seconds before signaling to the broker they have processed a message. This test ends very shortly after the Kafka messages are consumed and so the consumer did not signal that it had processed them, until I shortened the `auto.commit.interval.ms` to 100. I could have put a delay in the test but I would rather the test ends as soon as possible.

nslsii/__init__.py
```diff
@@ -548,20 +545,26 @@ def subscribe_kafka_publisher(RE, beamline_name, bootstrap_servers, producer_con
     producer_config: dict
         dictionary of Kafka Producer configuration settings
 
-    Returns
```
gwbischof, line 551: Maybe add the Returns section to the docstring, because this function returns `topic` now.

jklynch: done!

jklynch:
The latest CI failure is only on python 3.6.

jklynch:
That's a good question. I have not thought about it. In `test_publisher_with_no_broker` there should be no time penalty related to Kafka since the producer is never used.

So I ran three quick and dirty tests with `RE(count([det2]))`, `RE(count([det2]), num=10)`, and `RE(count([det2]), num=100)`. They ran in 5.01s, 5.03s, and 5.2s respectively. So this seems ok other than that there seems to be a 5s overhead.

I also ran those tests under `pyinstrument` and it says

```
Program: /home/jlynch/local/beamlines/rsoxs/nxsas-worker/venv/bin/pytest -s nslsii/tests/test_kafka_publisher.py

5.031 test_publisher_with_no_broker  test_kafka_publisher.py:131
└─ 5.031 __call__  bluesky/run_engine.py:700
      [6 frames hidden]  bluesky, threading
         5.031 wait  threading.py:270


time for count: 5.031
```
I think I'll have to talk to Tom or Dan to figure that out.

jklynch:
I will create an issue about performance. I would like to address it with a test if possible in a later PR.

jklynch:
After some more thought it seems like the performance issue should be tested in `bluesky-kafka`. What do you think @gwbischof ?

gwbischof:
Seems good to me, yea I think it makes sense to test the performance issue in bluesky-kafka.

gwbischof:
Last time the tests ran 3.6 failed, now 3.7 is failing. Am I correct about that?

gwbischof:
What do you think about also moving the other kafka tests to bluesky-kafka? Also subscribe_kafka_publisher might be useful to have in bluesky-kafka. That could be something that we do later.

jklynch:
I want to have a tests here for `subscribe_kafka_publisher` even though they are similar to the tests in `bluesky_kafka`. These tests can be more focused on how we work with Kafka at the beamlines. That's my intention anyway.

jklynch:
I ran the python 3.7 build again and it passed.

Issue #94, opened 2020-08-05
Xspress3Channel class


bruceravel:
I wish that the Xspress3Channel class in nslsii/detectors/xspress3.py defined an attribute like this near line 433
```
extra_rois_enabled = Cpt(EpicsSignal, 'PluginControlValExtraROI')
```
Seems silly to subclass just to add this obviously useful attribute.

mrakitin:
PRs are always welcome! Is it a standard part of the detector?

mrakitin:
We discussed it with Bruce today, and decided to give it a try at a couple of beamlines which use xspress3 (QAS and SRX maybe) to make sure it does not break what they are doing.

PR #95, opened 2020-08-11
Add a control to Xspress3Channel to enable/disable ROIs 5 through 16

bruceravel:
This adds an attribute to the Xspress3Channel class to toggle on ROIs 5 through 16.  The IOC seems to disable all ROIs by default. The CSS screen provides controls for enabling these, but save/restore for the standard IOC does not seem to remember this change.  I want to use the high numbered ROIs, so I need access to these PVs when I instantiate an instance of this class.

I think this changes costs very little and provides a sensible interface to a PV to which I need a sensible interface.

### DIFF
--- a/nslsii/detectors/xspress3.py
+++ b/nslsii/detectors/xspress3.py
@@ -431,7 +431,8 @@ class Xspress3Channel(ADBase):
 
     rois = DDC(make_rois(range(1, 17)))
     vis_enabled = Cpt(EpicsSignal, 'PluginControlVal')
-
+    extra_rois_enabled = Cpt(EpicsSignal, 'PluginControlValExtraROI')
+    
     def __init__(self, prefix, *, channel_num=None, **kwargs):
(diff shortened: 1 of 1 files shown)


bruceravel:
(See #94)

bruceravel:
I don't know the answers to your questions.  What I do know if that the high numbered ROIs started disabled and that this is the PV that enables them.  At least on the instance at BMM.  Hmmm... occurs to me that the Centos box that came with the new Xspress3 is running a fairly old version of AD.  I suppose it's possible that this has changed between then and now.  I know that Quantum is not yet supporting/distributing the extensive work out of GSECARS, which seems awfully lazy.

lcdesilva:
From what I see, this PV is not available in the EPICS IOC version from APS. I will have to take a closer look to confirm. For the Quantum IOC version, the PV for enabling 1-4 is different from 5-16. Just wanted to make that latter note.

Issue #96, opened 2020-08-14
Triggering of xspress3 at BMM takes 2.5s for 0.5s exposure


mrakitin:
This part of the code seems to be a bottleneck (see the attached picture with the `ts_msg_hook`-generated output):
https://github.com/NSLS-II/nslsii/blob/18c06329edb76f068d63e5b6522a6781d2bc696c/nslsii/detectors/xspress3.py#L623-L638

![20200813_155248](https://user-images.githubusercontent.com/13209176/90202011-737d6d00-ddaa-11ea-81cf-0a958534b06f.jpg)

mrakitin:
Did a few tests today with @bruceravel, all repeatable, even with the disabled HDF5 plugin.

@tacaswell, what would be the best way to debug it? Which kinds of prints can help here? I guess this dispatching is doing something very slowly, but not sure how to best approach it.

jklynch:
https://github.com/joerick/pyinstrument

tacaswell:
I think we may need to throw a whole lot of camonitors on the AD signals to see where it is getting stuck.  If we are at the wait message, than that means we are through `trigger` and are waiting now waiting for the code in https://github.com/NSLS-II/nslsii/blob/18c06329edb76f068d63e5b6522a6781d2bc696c/nslsii/detectors/xspress3.py#L615-L621 to fire.

PR #97, opened 2020-08-19
MNT: do not try to erase spectrum in trigger

tacaswell:
This has been observed to lock up the IOC for ~2s.

We applied this fix in the profile at BMM with @bruceravel

### DIFF
--- a/nslsii/detectors/xspress3.py
+++ b/nslsii/detectors/xspress3.py
@@ -148,6 +148,8 @@ class Xspress3FileStore(FileStorePluginBase, HDF5Plugin):
         self.settings.acquire.put(0, wait=True)
 
         total_points = self.parent.total_points.get()
+        if total_points < 1:
+            raise RuntimeError("You must set the total points")
         spec_per_point = self.parent.spectra_per_point.get()
         total_capture = total_points * spec_per_point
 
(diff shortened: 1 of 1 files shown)


tacaswell:
I also added a commit to make it clear that you need to set the total number of points before the device can be used.

mrakitin:
That's an unexpected source of the issue. Nice work on figuring it out!

mrakitin:
Can we claim it closes #96?

tacaswell:
Following discussion with @jwlodek it looks like pushing erase on every trigger is redundant.  The function that gets called internally allocates 0 arrays, pushes them through the AD pipeline and resets the histograms.  However, the function called when we push the acquire button also resets the histograms and pushes the newly collected data down epics pipe.

https://github.com/epics-modules/xspress3/blob/39832f64ff7f246b2f6dcb91b6ff63a7e8f8cbfd/xspress3App/src/xspress3Epics.cpp#L868

 https://github.com/epics-modules/xspress3/blob/39832f64ff7f246b2f6dcb91b6ff63a7e8f8cbfd/xspress3App/src/xspress3Epics.cpp#L1161

mrakitin:
I think this fix should go in before the next push. @tacaswell, are there any concerns left here?

tacaswell:
no, I don't think so.

tacaswell:
This may break things, but if it does it is because it was already broken we just did not know it!

mrakitin:
Thank you!

PR #100, opened 2020-08-24
Make TwoButtonShutter configurable.

danielballan:
FXI needs more retries. Sure wish we could get this fixed in the IOC!


danielballan:
One build had a random failure in the kafka publisher test, unrelated to this PR.

PR #101, opened 2020-08-26
handle exceptions from Kafka producers

jklynch:
This PR adds exception handling to Kafka message publishing. The intention is to insulate the RunEngine from Kafka failures.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -213,7 +213,12 @@ def configure_base(
             RE,
             beamline_name=broker_name,
             bootstrap_servers="cmb01:9092,cmb02:9092,cmb03:9092",
-            producer_config={"enable.idempotence": True},
+            producer_config={
+                "acks": 0,
+                "message.timeout.ms": 3000,
+                "queue.buffering.max.kbytes": 10 * 1048576,  # default is 1048576
+                "compression.codec": "snappy"
+            },
         )
 
     # always configure %xmode minimal
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -576,17 +576,19 @@ def subscribe_kafka_publisher(RE, beamline_name, bootstrap_servers, producer_con
         def handle_publisher_exceptions(name_, doc_):
             """
             Do not let exceptions from the Kafka producer kill the RunEngine.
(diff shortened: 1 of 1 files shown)


jklynch:
`queued.max.messages.kbytes` seems to be the Consumer parameter. I think the analogous Producer parameter is `queue.buffering.max.kbytes` so I'll set that to 10 times the default. I'll add compression as well.

nslsii/__init__.py
```diff
@@ -568,6 +573,21 @@ def kafka_publisher_factory(name, start_doc):
...
 
+        def handle_publisher_exceptions(name_, doc_):
+            """
+            Do not let exceptions from the Kafka producer kill the RunEngine.
+            """
+            try:
+                kafka_publisher(name_, doc_)
+            except Exception as ex:
+                logger = logging.getLogger("nslsii")
+                logger.error(
```
danielballan, line 584: `logger.exception` takes in a message string and _automatically_ includes the current traceback below it to create an ERROR-level log record. Example:

```py
In [3]: try: 
   ...:     1/0 
   ...: except Exception: 
   ...:     logger.exception("Just minding my own business when all of a sudden") 
   ...:                                                                                                                                                                                       
Just minding my own business when all of a sudden
Traceback (most recent call last):
  File "<ipython-input-3-c10aadf263f7>", line 2, in <module>
    1/0
ZeroDivisionError: division by zero
```

Thus, I think these two logger calls should be combined: drop `logger.error(...)` and pass its message to `logger.exception`.

gwbischof:
I tested this on cmb, a run with 5000 takes 15s, the same run with kafka subscribed takes 20s.
The compression option didn't affect the execution time.

gwbischof:
Do we need to do anything special with the consumers if we turn on compression?

jklynch:
I think the consumers know if compression is used.

nslsii/__init__.py
jklynch: combined these two and two more farther down

mrakitin (approved):
Looks good to me.

mrakitin:
@jklynch said the tests pass locally. Going to merge it now.

PR #102, opened 2020-08-27
.value --> .get()

mrakitin:
Observed during SRX deployment (Part 3) (https://github.com/NSLS-II-SRX/profile_collection/issues/10).

### DIFF
--- a/nslsii/iocs/thermo_sim.py
+++ b/nslsii/iocs/thermo_sim.py
@@ -52,7 +52,7 @@ class Thermo(PVGroup):
                     setpoint)
 
         T = t_rbv(T0=self._T0,
-                  **{k: getattr(self, k).get()
+                  **{k: getattr(self, k).value
                      for k in ['setpoint', 'K', 'omega', 'Tvar']})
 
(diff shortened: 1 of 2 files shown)


mrakitin:
Thanks for reviewing this, @danielballan! Do you mind to check https://github.com/NSLS-II-HXN/hxntools/pull/19 as well? Identical updates.

nslsii/iocs/thermo_sim.py
```diff
@@ -52,7 +52,7 @@ def t_rbv(T0, setpoint, K, omega, Tvar,):
                     setpoint)
 
         T = t_rbv(T0=self._T0,
-                  **{k: getattr(self, k).value
```
tacaswell, line 55: This is inside of a caproto IOC and I don't think will work!

mrakitin: Reverted this change. Thanks for catching it up!

mrakitin:
@tacaswell, are you fine with the changes now? Thanks.

PR #104, opened 2020-09-02
Sync useful changes with hxntools

mrakitin:
I haven't added the `.value` -> `.get()` changes as they are covered by #102. I ignored the old parts of code from `hxntools` which were fixed here already (e.g., #97).

### Organizational note:
I think we should keep the `xspress3`-related code here in this repo, and remove it from `hxntools`, so we don't have to guess where to put future updates.


mrakitin:
Thanks Tom!

PR #105, opened 2020-09-04
release notes for v0.1.0

jklynch:


mrakitin (approved):
Thanks, looks good to me. I wonder if we should tag it as v0.1.0. Were there enough changes?

jklynch:
I was also thinking v0.1.0 seems reasonable.

docs/source/release-history.rst
```diff
@@ -2,6 +2,13 @@
 Release History
 ***************
 
+v0.1.0 (2020-09-04)
+====================
```
mrakitin, line 6: I think it's too many equal signs here.
```suggestion
===================
```

mrakitin (commented):
Looks good, just need to make sphinx happy.

mrakitin (approved):

PR #109, opened 2020-10-23
Fix xspress3 status

mrakitin:
This is a follow-up PR for the fix we came up with @tacaswell and @bruceravel on a call today. This will eliminate the error message coming from the background thread on the second count of the xspress3 detector object.


tacaswell (approved):

tacaswell:
Merge at-will when CI is green.

mrakitin:
Thanks for your review, @tacaswell!

PR #112, opened 2020-12-01
set propagate = False on loggers

jklynch:
This PR proposes a solution to the problem of users being overwhelmed with log output from the loggers configured in `nslsii.configure_base` when they configure their own root logger.

By setting `getLogger(["bluesky", "caproto", "ophyd", "nslsii"]).propagate = False`, log messages sent to those loggers will not propagate to higher-level loggers such as a root logger. This will allow our logging to proceed without disturbing users who configure a root logger to print their log messages to the console.

The documentation for `nslsii.configure_bluesky_logging` has been updated to reflect this change and to direct users to set `getLogger(...).propagate = True` in their code if they want log messages to propagate from our loggers to higher-level loggers.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -343,24 +343,10 @@ def configure_bluesky_logging(ipython, appdirs_appname="bluesky", propagate_log_
     )
     log_file_handler.setFormatter(logging.Formatter(fmt=log_file_format))
 
-    logging.getLogger("bluesky").addHandler(log_file_handler)
-    logging.getLogger("caproto").addHandler(log_file_handler)
-    logging.getLogger("ophyd").addHandler(log_file_handler)
-    logging.getLogger("nslsii").addHandler(log_file_handler)
-
-    # set the loggers to send INFO and higher log
-    # messages to their handlers
-    logging.getLogger("bluesky").setLevel("INFO")
-    logging.getLogger("caproto").setLevel("INFO")
-    logging.getLogger("ophyd").setLevel("INFO")
-    logging.getLogger("nslsii").setLevel("INFO")
-
-    # configure loggers so that log messages do not
-    # propagate to higher level loggers
-    logging.getLogger("bluesky").propagate = propagate_log_messages
-    logging.getLogger("caproto").propagate = propagate_log_messages
-    logging.getLogger("ophyd").propagate = propagate_log_messages
-    logging.getLogger("nslsii").propagate = propagate_log_messages
+    for logger_name in ("bluesky", "caproto", "ophyd", "nslsii"):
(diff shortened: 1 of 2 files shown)


gwbischof (approved):
Looks good!

mrakitin (commented):
I have small suggestions.

nslsii/__init__.py
```diff
@@ -336,28 +342,30 @@ def configure_bluesky_logging(ipython, appdirs_appname="bluesky"):
...
     # configure loggers so that log messages do not
     # propagate to higher level loggers
-    logging.getLogger("bluesky").propagate = False
-    logging.getLogger("caproto").propagate = False
-    logging.getLogger("ophyd").propagate = False
-    logging.getLogger("nslsii").propagate = False
+    logging.getLogger("bluesky").propagate = propagate_log_messages
+    logging.getLogger("caproto").propagate = propagate_log_messages
+    logging.getLogger("ophyd").propagate = propagate_log_messages
+    logging.getLogger("nslsii").propagate = propagate_log_messages
```
mrakitin, line 363: It feels a for-loop for will have with the repeated patterns at least in 3 places above.

mrakitin (commented):
Looks good. I suggested a small optimization.

nslsii/__init__.py
```diff
@@ -343,24 +343,10 @@ def configure_bluesky_logging(ipython, appdirs_appname="bluesky", propagate_log_
...
-    # configure loggers so that log messages do not
-    # propagate to higher level loggers
-    logging.getLogger("bluesky").propagate = propagate_log_messages
-    logging.getLogger("caproto").propagate = propagate_log_messages
-    logging.getLogger("ophyd").propagate = propagate_log_messages
-    logging.getLogger("nslsii").propagate = propagate_log_messages
+    for logger_name in ("bluesky", "caproto", "ophyd", "nslsii"):
+        logging.getLogger(logger_name).addHandler(log_file_handler)
+        logging.getLogger(logger_name).setLevel("INFO")
+        logging.getLogger(logger_name).propagate = propagate_log_messages
```
mrakitin, line 349: `logging.getLogger(logger_name)` is repeated for 3 times above, so maybe another small optimization can be done, such as:
```py
    for logger_name in ("bluesky", "caproto", "ophyd", "nslsii"):
        logger = logging.getLogger(logger_name)
        logger.addHandler(log_file_handler)
        logger.setLevel("INFO")
        logger.propagate = propagate_log_messages
```

But I am happy with the current state of things too.

mrakitin (approved):
Looks good to me. An optional improvement is suggested, but it's good as is now.

mrakitin (approved):
Thanks for addressing all my suggestions, @jklynch!

PR #116, opened 2021-01-26
FIX: be forgiving of type instability in TwoButtonShutter status PV

tacaswell:
# This needs to be on floor at every beamline by noon tomorrow.

------

https://github.com/bluesky/ophyd/pull/946 changed it so that the value
passed to the callabck is the str version of the value not the
integer.

This means that the conversion of int -> str we are doing in the
callbacks of TwoButtonShutter are redundant and failing.  This means
that we will not ever mark the status object as done (and tend to fail
with a miss leading error due to trying to set the shutter again on
the way out of the RE).

This is a modified version of https://github.com/NSLS-II-ISS/profile_collection/pull/6 which is reportedly working at ISS

### DIFF
--- a/nslsii/devices.py
+++ b/nslsii/devices.py
@@ -51,7 +51,7 @@ class TwoButtonShutter(Device):
         def shutter_cb(value, timestamp, **kwargs):
             try:
                 value = enums[int(value)]
-            except ValueError:
+            except (ValueError, TypeError):
                 # we are here because value is a str not int
                 # just move on
                 ...
@@ -67,7 +67,7 @@ class TwoButtonShutter(Device):
             nonlocal count
             try:
                 value = cmd_enums[int(value)]
-            except ValueError:
+            except (ValueError, TypeError):
                 # we are here because value is a str not int
                 # just move on
                 ...
--- a/nslsii/devices.py
+++ b/nslsii/devices.py
@@ -97,7 +97,7 @@ class TwoButtonShutter(Device):
 
         return st
 
-    def stop(self, success):
+    def stop(self, *, success=False):
         import time
         prev_st = self._set_st
         if prev_st is not None:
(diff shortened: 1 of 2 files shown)


mrakitin (approved):
Thanks for the quick and important fix, @tacaswell !

nslsii/iocs/eps_two_state_ioc_sim.py
```diff
@@ -2,6 +2,25 @@
...
 from caproto.server import template_arg_parser, run
 from caproto import ChannelType
+import contextvars
+import functools
+
+internal_process = contextvars.ContextVar('internal_process',
+                                          default=False)
+
+
+def no_reentry(func):
```
danielballan, line 12: I recall some discussion in caproto that, IIRC, this does not do what you intend. But since I could be wrong and this is just a IOC supporting a test, I will not hold up the PR over it.

nslsii/devices.py
```diff
@@ -49,7 +49,12 @@ def set(self, val):
         enums = self.status.enum_strs
 
         def shutter_cb(value, timestamp, **kwargs):
-            value = enums[int(value)]
+            try:
+                value = enums[int(value)]
+            except (ValueError, TypeError):
```
danielballan, line 54: Why so permissive? Should we check that the value is one of the expected enum types?

tacaswell: Because I want to avoid having more than one cycle of getting a weighing version to the floor, we can make it stricter later.

nslsii/iocs/eps_two_state_ioc_sim.py
tacaswell: This prevents the validator from being entered a second time in the same task which is exactly what we want here (or the internal write will infinitely recourse).

danielballan: I think "task" is not what you think it is here, but I'll pick this up later.

PR #118, opened 2021-03-25
use new kafka cluster

gwbischof:


nslsii/__init__.py
```diff
@@ -219,7 +219,7 @@ def configure_base(
         subscribe_kafka_publisher(
             RE,
             beamline_name=broker_name,
-            bootstrap_servers="cmb01:9092,cmb02:9092,cmb03:9092",
+            bootstrap_servers="kafka1:9092,kafka2:9092,kafka3:9092,kafka4:9092,kafka5:9092,kafka6:9092,kafka7:9092",
```
mrakitin, line 222: Would this work?
```py
In [1]: ",".join([f"kafka{i}:9092" for i in range(1, 7+1)])
Out[1]: 'kafka1:9092,kafka2:9092,kafka3:9092,kafka4:9092,kafka5:9092,kafka6:9092,kafka7:9092'
```

It's a bit less error-prone. In case a server is removed (i.e., the sequence is broken), we can replace `range(1, 8)` with the explicit list (e.g., `[1, 3, 5, 7]`).

mrakitin (commented):
Thanks, @gwbischof. I haven't touched the `kafka*` servers, so I rely on you in this case.

Below is a minor enhancement suggestion.

jklynch:
As @danielballan pointed out on slack, using an environment variable to configure the Kafka bootstrap servers is going to be important. This might be the right time to implement something like that.

jklynch (approved):
In the future we may want a second way to set `bootstrap_servers` but right now this seems sufficient.

mrakitin (approved):

PR #122, opened 2021-07-13
Publish kafka messages from a thread

jklynch:
This PR modifies the function that subscribes a Kafka Publisher to the RunEngine by inserting a thread between the two. The RunEngine places documents in a queue, and the "publishing thread" takes those documents off the queue and publishes them. The intention is to insulate plans from Kafka exceptions using the same idea as is used for olog.

Significant improvements have been made to the test coverage.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -216,7 +216,7 @@ def configure_base(
         configure_ipython_logging(exception_logger=log_exception, ipython=ipython)
 
     if publish_documents_to_kafka:
-        subscribe_kafka_publisher(
+        _build_and_subscribe_kafka_publisher(
             RE,
             beamline_name=broker_name,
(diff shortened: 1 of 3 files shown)


nslsii/__init__.py
```diff
@@ -547,11 +547,96 @@ def migrate_metadata():
     new_md.update(old_md)
 
 
-def subscribe_kafka_publisher(RE, beamline_name, bootstrap_servers, producer_config):
+def subscribe_kafka_publisher(RE, publisher_queue, kafka_publisher):
```
tacaswell, line 550: Slightly concerned about the API change.

jklynch: This should have been named to indicate it's not for public consumption. I have made that change.
Also I am confident this function is used nowhere else at present.

tacaswell (approved):

PR #126, opened 2021-07-30
send logging output to syslog in addition to log files

jklynch:
This PR adds a SysLogHandler to the various loggers.


mrakitin (approved):
Looks like `bluesky` does not need any special permissions to write to the system log, which is great :). Looks good to me, eager to try on a beamline!

PR #146, opened 2022-01-25
ENH: add SRX MAIA code

tacaswell:
This is resurrecting code we used at SRX for use at XFM.


mrakitin (commented):
Probably need to declare `recordwhat` in `requirements.txt`. Otherwise looks good to me.

nslsii/detectors/maia.py
```diff
@@ -0,0 +1,873 @@
...
+import time as ttime
+
+import numpy as np
+
+from ophyd import Device, Component as FldCpt, EpicsSignal, EpicsSignalRO
+from ophyd.status import DeviceStatus
+from ophyd.signal import Signal
+
+
+from recordwhat import RecordBase, _register_record_type, FieldComponent as Cpt
```
klauer, line 11: Mistaken `as Cpt` here instead of above?

```suggestion
from recordwhat import RecordBase, _register_record_type, FieldComponent as FldCpt
```

I'd also just wholesale nab `AiRecord` and all from recordwhat if I were you 🤷‍♂️

tacaswell:
I'm not sure it we want to merge this or not. 

It may be possible to simplify this and/or drop the record what dependency as Ken suggests.

This and record what are currently in an overlay directory at xfm

nslsii/detectors/maia.py
tacaswell: 🤦🏻 fixed.

gwbischof:
Rebased it on master

mrakitin:
Is it ready for review? We need it on the floor shortly.

mrakitin:
I am going to merge it as the XFM beamline runs the MAIA code from this PR branch.

PR #147, opened 2022-02-18
correct docstring for configure_base

jklynch:
The default value for parameter `publish_documents_with_kafka` was incorrectly reported as 'True'. This PR changes it to the correct value `False`.


mrakitin (approved):
What an oversight! Thanks for the fix, @jklynch!

PR #153, opened 2022-06-03
add utilities for running commands and folder permission checking

JunAishima:
* from py4xs.utils and lix_profile_collection/03-security,
   respectively
 * `check_access` was requested by Maksim, `run` looked useful to me

### DIFF
--- a/nslsii/utils.py
+++ b/nslsii/utils.py
@@ -1,0 +1,40 @@
+import os
+import re
+import subprocess
+
+
+def run(cmd, path="", ignoreErrors=True, returnError=False, debug=False):
+    """cmd should be a list, e.g. ["ls", "-lh"]
+    path is for the cmd, not the same as cwd
+    """
+    cmd[0] = path + cmd[0]
+    p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    out, err = p.communicate()
+    if debug:
+        print(out.decode(), err.decode())
+    if len(err) > 0 and not ignoreErrors:
+        print(err.decode())
+        raise Exception(err.decode())
+    if returnError:
+        return out.decode(), err.decode()
+    else:
+        return out.decode()
+
+
+def check_access(fn):
+    if not os.path.exists(fn):
+        raise Exception(f"{fn} does not exist ...")
+    if os.access(fn, os.W_OK):
+        print(f"write access to {fn} verified ...")
+        return
+
+    # this below may not be necessary
+    out = run(["getfacl", "-cn", fn])
+    wgrps = [int(t[:-4].lstrip("group:")) for t in re.findall("groups:[0-9]*:rw.", out)]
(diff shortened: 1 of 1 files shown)


nslsii/utils.py
```diff
@@ -0,0 +1,40 @@
+import os
+import re
+import subprocess
+
+
+def run(cmd, path="", ignoreErrors=True, returnError=False, debug=False):
```
mrakitin, line 6: I'd prefer to name the kwargs in the snake_case style, i.e.:
- `ignoreErrors` -> `ignore_errors`
- `returnError` -> `return_error`

nslsii/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    """cmd should be a list, e.g. ["ls", "-lh"]
```
mrakitin, line 7: The docstring should have a break between the first line and the following lines. Maybe an introductory sentence about this function can be useful.

Also, `run` seems to be too vague. Maybe something more specific such as `execute_command` can be more appropriate.

nslsii/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    out, err = p.communicate()
+    if debug:
+        print(out.decode(), err.decode())
+    if len(err) > 0 and not ignoreErrors:
+        print(err.decode())
+        raise Exception(err.decode())
+    if returnError:
+        return out.decode(), err.decode()
+    else:
+        return out.decode()
```
mrakitin, line 21: `out.decode()` and `err.decode()` are used a few times. Maybe worth calling them once, and reuse via variables?

nslsii/utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def check_access(fn):
+    if not os.path.exists(fn):
+        raise Exception(f"{fn} does not exist ...")
+    if os.access(fn, os.W_OK):
+        print(f"write access to {fn} verified ...")
```
mrakitin, line 28: For this message, I think it will make sense to add a few words about this being verified via Unix permissions.

nslsii/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    # this below may not be necessary
+    out = run(["getfacl", "-cn", fn])
+    wgrps = [int(t[:-4].lstrip("group:")) for t in re.findall("groups:[0-9]*:rw.", out)]
+    ugrps = os.getgroups()
+    if len(set(wgrps) & set(ugrps)) == 0:
+        print("groups with write permission: ", wgrps)
+        print("user group membership: ", ugrps)
+        raise Exception(f"the current user does not have write access to {fn}")
+    else:
+        print(f"write access to {fn} verified ...")
```
mrakitin, line 40: For this message, I think it will make sense to add a few words about this being verified via ACL (`getfacl`).

mrakitin (commented):
Thanks for porting the tools. The code needs a little work to make it more compatible with the rest of the code base.

PR #156, opened 2022-07-05
ENH: give more control over where the kafka documents are published

tacaswell:


tacaswell:
https://github.com/NSLS-II-PDF/mmm-experiments/blob/main/mmm_experiments/data/_kafka_publisher.py 

I vendored these changes out for the MMM experiment at PDF.  This PR does not reflect the changes to where this helper gets merged.

I do not think it hurts to merge this as-is, but it is not complete.

jklynch:
This should maybe go to bluesky-kafka.

PR #157, opened 2022-07-13
Improvements to RunEngineRedisDict

jklynch:
The original implementation of nslsii.md_dict.RunEngineRedisDict failed to maintain data types for global metadata. This PR proposes a fix, improves the comments, refactors some code for better readability, and expands the tests.

The "fix" adds a dictionary to map global metadata keys to types.

### DIFF
--- a/nslsii/md_dict.py
+++ b/nslsii/md_dict.py
@@ -1,6 +1,6 @@
 import logging
+import re
 from collections import ChainMap, UserDict
-from pprint import pformat
 from uuid import uuid4
 
 import msgpack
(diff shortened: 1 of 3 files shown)


nslsii/tests/conftest.py
```diff
@@ -27,8 +27,16 @@ def redis_dict_factory():
...
-    def _factory(re_md_channel_name):
-        return RunEngineRedisDict(host="localhost", port=6379, db=0, re_md_channel_name=re_md_channel_name)
+    def _factory(re_md_channel_name, **kwargs):
+        init_kwargs = {
+            "host": "localhost",
+            "port": 6379,
+            "db": 0,
+            "re_md_channel_name": re_md_channel_name
+        }
+        init_kwargs.update(kwargs)
```
danielballan, line 37: Should this use `setdefault` so that `kwargs` get precedence?

nslsii/tests/test_redis_dict.py
```diff
@@ -58,6 +58,27 @@ def test_instantiate_no_server():
         RunEngineRedisDict(host="localhost", port=9999)
 
 
+def test__parse_message_data():
```
danielballan, line 61: Accidentally double underscore

nslsii/md_dict.py
```diff
@@ -14,15 +14,39 @@
...
+         The drawback is that this "local" metadata key-values are not
+         directly readable or writeable by Redis clients.
+    """
+
     PACKED_RUNENGINE_METADATA_KEY = "runengine-metadata-blob"
 
+    # these are the values used for global metadata
+    # if it does not exist when the RunEngineRedisDict
+    # is created
+    _missing_value = {str: "", int: int(-1), float: -1.0}
```
danielballan, line 40: Would it be better to allow the expected types to be `... or None` so that a missing value is an unambiguous out-of-band sentinel, rather than an in-band magic value? The branching logic that this leads to below looks potentially confusing. And it could it result in a field like `scan_id` being assigned `-1`?

danielballan (changes requested):

nslsii/tests/test_redis_dict.py
jklynch: Actually intentional double underscore, since the method under test starts with "_".

nslsii/md_dict.py
jklynch: There were some complications here, and I landed on what seemed like an ok solution. Using msgpack with the "local" metadata makes this easy, but the "global" metadata is more difficult because those are pure Redis key-values. I think revisiting it now is a good idea.

Maybe a better solution than "default" global metadata values is just no default for those keys. Then RE.md raises KeyError and the user is made aware of the problem, which should happen very rarely.

danielballan (approved):

PR #159, opened 2022-08-02
Initial pass at creating ophyd classes for the QEPro spectrometer IOC

jwlodek:
Adds a class for the QEPro IOC I have developed for XPD. For now added it under `detectors` since I'm not sure what the best place for it is.

### DIFF
--- a/nslsii/detectors/QEPro.py
+++ b/nslsii/detectors/QEPro.py
@@ -1,0 +1,135 @@
+import logging
+
+from ophyd import (Device, Component as Cpt, FormattedComponent as FC,
+                   Signal)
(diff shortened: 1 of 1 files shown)


nslsii/detectors/QEPro.py
```diff
@@ -0,0 +1,166 @@
...
+        self.get_reference_frame()
+
+
+    def trigger(self):
+
+        self.acquire.put(1, wait=True)
+
+    
+
+
```
mrakitin, line 166:
```suggestion
```

mrakitin (approved):
Looks good to me. The style can be checked/corrected with flake8/black, but it can go in for the upcoming release "as is" (we can apply the formatting for the whole code base later).

PR #162, opened 2022-10-13
Add a helper warmup function for hdf5 plugins of AD

mrakitin:
This was originally implemented at [NSLS-II TES](https://github.com/NSLS-II-TES/profile_collection/blob/d8096818588285185ca2837d5988ec8f62fd5e1c/startup/00-startup.py#L110-L128), porting the function to here to be available to other beamlines.

### DIFF
--- a/nslsii/detectors/utils.py
+++ b/nslsii/detectors/utils.py
@@ -61,3 +61,29 @@ def make_filename_add_subdirectory(fn, read_path, write_path, *,
     if make_directories:
         makedirs(read_path)
     return fn, read_path, write_path
+
+
+def warmup_hdf5_plugins(detectors):
+    """Warm-up the hdf5 plugins for a list of detectors.
+
+    This is necessary for when the corresponding IOC restarts, we have to
+    trigger one image for the hdf5 plugin to work correctly, else we get file
+    writing errors.
+
+    Parameters:
+    -----------
+    detectors : list
+        a list of detector ophyd objects which have the .hdf5 plugins
+    """
+    for det in detectors:
+        if hasattr(det, "hdf5"):
+            _array_size = det.hdf5.array_size.get()
+            if 0 in [_array_size.height, _array_size.width]:
+                print(f"\n  Warming up HDF5 plugin for {det.name} as the "
+                      f"array_size={_array_size}...")
+                det.hdf5.warmup()
+                print(f"  Warming up HDF5 plugin for {det.name} is done. "
+                      f"array_size={det.hdf5.array_size.get()}\n")
+            else:
+                print(f"\n  Warming up of the HDF5 plugin is not needed for "
+                      f"{det.name} as the array_size={_array_size}.")


prjemian:
Since this method is needed for other file area detector writing plugins, perhaps it should be more general?  Here is similar code from `apstools` with such generality:

```python
def AD_plugin_primed(plugin):
    """
    Has area detector pushed an NDarray to the file writer plugin?  True or False

    PARAMETERS

    plugin
        *obj* :
        area detector plugin to be *primed* (such as ``detector.hdf1``)

    EXAMPLE::

        AD_plugin_primed(detector.hdf1)

    Works around an observed issue: #598
    https://github.com/NSLS-II/ophyd/issues/598#issuecomment-414311372

    If detector IOC has just been started and has not yet taken an image
    with the file writer plugin, then a TimeoutError will occur as the
    file writer plugin "Capture" is set to 1 (Start).  In such case,
    first acquire at least one image with the file writer plugin enabled.

    Also issue in apstools (needs a robust method to detect if primed):
    https://github.com/BCDA-APS/apstools/issues/464

    Since Area Detector release 2.1 (2014-10-14).

    The *prime* process is not needed if you select the
    *LazyOpen* feature with *Stream* mode for the file plugin.
    *LazyOpen* defers file creation until the first frame arrives
    in the plugin. This removes the need to initialize the plugin
    with a dummy frame before starting capture.
    """
    cam = plugin.parent.cam
    tests = []


    for obj in (cam, plugin):
        test = np.array(obj.array_size.get()).sum() != 0
        tests.append(test)
        if not test:
            logger.debug("'%s' image size is zero", obj.name)


    checks = dict(
        array_size=False,
        color_mode=True,
        data_type=True,
    )
    for key, as_string in checks.items():
        c = getattr(cam, key).get(as_string=as_string)
        p = getattr(plugin, key).get(as_string=as_string)
        test = c == p
        tests.append(test)
        if not test:
            logger.debug("%s does not match", key)

    return False not in tests
```

prjemian:
Here's the [code that primes the plugin](https://github.com/BCDA-APS/apstools/blob/97970d180c8b9b8a6119ff12c5505884f9b63bac/apstools/devices/area_detector_support.py#L199-L249):

```python
def AD_prime_plugin2(plugin):
    """
    Prime this area detector's file writer plugin.

    Collect and push an NDarray to the file writer plugin.
    Works with all file writer plugins.

    Based on ``ophyd.areadetector.plugins.HDF5Plugin.warmup()``.

    PARAMETERS

    plugin
        *obj* :
        area detector plugin to be *primed* (such as ``detector.hdf1``)

    EXAMPLE::

        AD_prime_plugin2(detector.hdf1)

    """
    if AD_plugin_primed(plugin):
        logger.debug("'%s' plugin is already primed", plugin.name)
        return


    sigs = OrderedDict(
        [
            (plugin.enable, 1),
            (plugin.parent.cam.array_callbacks, 1),  # set by number
            (plugin.parent.cam.image_mode, 0),  # Single, set by number
            # Trigger mode names are not identical for every camera.
            # Assume here that the first item in the list is
            # the best default choice to prime the plugin.
            (plugin.parent.cam.trigger_mode, 0),  # set by number
            # just in case the acquisition time is set very long...
            (plugin.parent.cam.acquire_time, 1),
            (plugin.parent.cam.acquire_period, 1),
            (plugin.parent.cam.acquire, 1),  # set by number
        ]
    )


    original_vals = {sig: sig.get() for sig in sigs}


    for sig, val in sigs.items():
        time.sleep(0.1)  # abundance of caution
        sig.set(val).wait()


    time.sleep(2)  # wait for acquisition


    for sig, val in reversed(list(original_vals.items())):
        time.sleep(0.1)
        sig.set(val).wait()
```

jklynch (changes requested):
I would like to make 2 recommendations:
  1. this code should go in a new file nslsii/areadetector/utils.py
  2. Pete Jemian's code should also go there and be used to implement warmup_hdf5_plugins

tacaswell:
Can we please do this as a plan so it will compose nicely?

```python

def AD_prime_plugin2(plugin):
    """
    Prime this area detector's file writer plugin.

    Collect and push an NDarray to the file writer plugin.
    Works with all file writer plugins.

    Based on ``ophyd.areadetector.plugins.HDF5Plugin.warmup()``.

    PARAMETERS

    plugin
        *obj* :
        area detector plugin to be *primed* (such as ``detector.hdf1``)

    EXAMPLE::

        AD_prime_plugin2(detector.hdf1)

    """
    if AD_plugin_primed(plugin):
        logger.debug("'%s' plugin is already primed", plugin.name)
        return


    sigs = OrderedDict(
        [
            (plugin.enable, 1),
            (plugin.parent.cam.array_callbacks, 1),  # set by number
            (plugin.parent.cam.image_mode, 0),  # Single, set by number
            # Trigger mode names are not identical for every camera.
            # Assume here that the first item in the list is
            # the best default choice to prime the plugin.
            (plugin.parent.cam.trigger_mode, 0),  # set by number
            # just in case the acquisition time is set very long...
            (plugin.parent.cam.acquire_time, 1),
            (plugin.parent.cam.acquire_period, 1),
            (plugin.parent.cam.acquire, 1),  # set by number
        ]
    )

    original_vals = {}
    for sig in sigs:
        original_vals[sig] = yield from bps.rd(sig)

    for sig, val in sigs.items():
        yield from bps.sleep(.1)
        yield from bps.abs_set(sig, val)
        
    yield from bps.sleep(2)  # wait for acquisition


    for sig, val in reversed(original_vals.items()):
        yield from bps.sleep(.1)
        yield from bps.abs_set(sig, val)        

```

mrakitin:
We should revive it. Should it go to ophyd?

PR #163, opened 2022-10-14
Make naming and case of global keys be consistent with each other.

stuartcampbell:
Remove the need to have special knowledge that certain keys are upper case (i.e. `SAF`)
Changed SAF ID to match the naming scheme used for Proposal ID (i.e. `proposal_id`)


jklynch (approved):
This seems reasonable.

mrakitin (approved):

PR #166, opened 2023-01-09
Fix mis-use of prefix

jklynch:
This PR makes a small but important correction to the way HDF5Plugins were initialized in tests. The `prefix=` keyword argument should not have been used as it was. It has been replaced with the proper positional argument.


mrakitin (approved):
Looks good to me, thanks for the fix!

PR #167, opened 2023-01-17
Add dtype_str parameter to Xspress3ExternalFileReference.__init__

jklynch:
This PR proposes an additional parameter, `dtype_str`, for `Xspress3ExternalFileReference.__init__`. In addition a member variable is added to hold the `dtype_str` value and use it to when `describe()` is called. This allows user code to modify the array data type, which is not always `uint32` as was discovered at BMM.

### DIFF
--- a/nslsii/areadetector/xspress3.py
+++ b/nslsii/areadetector/xspress3.py
@@ -122,10 +122,23 @@ class Xspress3Trigger(Device):
 
 
 class Xspress3ExternalFileReference(Signal):
-    """ A special Signal for datum document information. """
+    """ A special Signal for datum document information.
 
-    def __init__(self, *args, bin_count=4096, dim_name="bin_count", **kwargs):
+    Parameters
+    ----------
+    dtype_str: str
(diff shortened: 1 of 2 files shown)


tacaswell (approved):

nslsii/areadetector/xspress3.py
```diff
@@ -135,7 +148,7 @@ def describe(self):
             dict(
                 external="FILESTORE:",
                 dtype="array",
-                dtype_str="uint32",  # <u4
+                dtype_str=self.dtype_str,
```
mrakitin, line 151: I remember for CHX the valid values for `dtype_str` were specified. Should we do this validation here too?

https://github.com/NSLS-II-CHX/profile_collection/blob/a33842a933d2db656b1f91178e7bb5ad8fdf239b/startup/20-area-detectors.py#L46-L49

jklynch: Once we know the valid values....

PR #170, opened 2023-06-02
Adding webcam/vstream module

jwlodek:


mrakitin (approved):
Tested with a test setup at PDF via https://github.com/mrakitin/profile_vstream/pull/1.

PR #171, opened 2023-06-02
Added optional call_returns_result parameter

RobertSchaffer1:


RobertSchaffer1:
@mrakitin I think this should do it?

mrakitin (approved):
Thanks for this useful addition, @RobertSchaffer1!
For reference: https://blueskyproject.io/bluesky/run_engine_api.html#the-runengineresult.

PR #174, opened 2024-01-03
feat: add pmac kill signal and switch

maffettone:
In response to https://github.com/NSLS-II-CSX/profile_collection/issues/84. 

Initially, an EPICS record was added to include a custom PV for Kill Status that would grab the 11th bit from the status. This moves that logic to the ophyd level, and constructs an appropriate switch that can kill a PMAC and check if the PMAC is killed. 

## TODO: 
- [x] Confirm that the naming approach is appropriate and broadly applicable.

### DIFF
--- a/nslsii/motors/delta_tau.py
+++ b/nslsii/motors/delta_tau.py
@@ -23,8 +23,8 @@ class PMACStatus(EpicsSignalRO):
 class PMACKillSwitch(Device):
     """
     A device to kill the PMAC-PC controller.
-    In the named version of the EPICS record, the status is expected at Sts:4-Sts.
-    Whereas in the axis numbered version, the status is expected at Sts:1-Sts.
+    In the named version of the EPICS record (e.g., XF:23ID1-ES{Dif-Ax:Del}), the status is expected at Sts:4-Sts.
+    Whereas in the axis numbered version (e.g., XF:23ID1-CT{MC:12-Ax:1}), the status is expected at Sts:1-Sts.
     """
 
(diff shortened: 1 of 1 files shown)


nslsii/motors/delta_tau.py
```diff
@@ -0,0 +1,45 @@
...
+class PMACKillSwitch(Device):
+    """
+    A device to kill the PMAC-PC controller.
+    In the named version of the EPICS record (e.g., XF:23ID1-ES{Dif-Ax:Del}), the status is expected at Sts:4-Sts.
+    Whereas in the axis numbered version (e.g., XF:23ID1-CT{MC:12-Ax:1}), the status is expected at Sts:1-Sts.
+    """
+
+    kill = Cpt(EpicsSignal, "Cmd:Kill-Cmd.PROC", kind=Kind.omitted)
+    status = Cpt(PMACStatus, "Sts:4-Sts", kind=Kind.normal)
+
```
jwlodek, line 29: The naming convention here afaik is actually based on the axis number. So for axis one, it would be Sts:1-Sts. For axis 2 it would be Sts:2-Sts and so on. So probably axis number should be an argument passed in at `__init__` and the component should be generated based on that.

jwlodek (changes requested):

jwlodek (approved):

johnsinsheimer:
Yes this is the correct naming scheme from the database:
motorstatus.db
There is kind of duplicate information / PV names in another record, axis status by number , but here it matches the "Axis Status by Name" menu

PR #175, opened 2024-01-31
Use correct number of dimension names in Xspress3ExternalFileReference

jmaruland:
This PR fixes an error that was identified in https://github.com/bluesky/databroker/pull/793 during tests for the Data Security project at SRX.
The error was tracked from tiled, back to databoker and last here where the default information of the dimensions of an Xspress3ExternalFileReference was being saved in a BlueSky document and this was causing an error while using this information and data to create an array in an xarray.Dataset.
The definition of `dim_name` was modified which considers the correct dimensions of the array and fixes the error when xarrays are generated in databroker.
The original use of `dim_name` was kept for cases where backwards compatibility is necessary.

### DIFF
--- a/nslsii/areadetector/xspress3.py
+++ b/nslsii/areadetector/xspress3.py
@@ -136,11 +136,14 @@ class Xspress3ExternalFileReference(Signal):
 
     """
 
-    def __init__(self, *args, dtype_str="uint32", bin_count=4096, dim_name="bin_count", **kwargs):
+    def __init__(self, *args, dtype_str="uint32", bin_count=4096, dim_name=["frame_number", "number_channels", "bin_count"], **kwargs):
         super().__init__(*args, **kwargs)
         self.dtype_str = dtype_str
         self.shape = (bin_count,)
-        self.dims = (dim_name,)
+        if isinstance(dim_name, str):
+            self.dims = (dim_name,) # Keeping this option for backwards compatibility with some beamlines
+        else:
+            self.dims = (*dim_name,)
 
     def describe(self):
         res = super().describe()
--- a/nslsii/areadetector/xspress3.py
+++ b/nslsii/areadetector/xspress3.py
@@ -141,7 +141,7 @@ class Xspress3ExternalFileReference(Signal):
         self.dtype_str = dtype_str
         self.shape = (bin_count,)
         if isinstance(dim_name, str):
-            self.dims = (dim_name,) # Keeping this option for backwards compatibility with some beamlines
+            self.dims = (dim_name,)  # Keeping this option for backward compatibility with some beamlines
         elif isinstance(dim_name, list):
             self.dims = (*dim_name,)
         else:
--- a/nslsii/areadetector/xspress3.py
+++ b/nslsii/areadetector/xspress3.py
@@ -136,14 +136,16 @@ class Xspress3ExternalFileReference(Signal):
 
     """
 
-    def __init__(self, *args, dtype_str="uint32", bin_count=4096, dim_name=["frame_number", "number_channels", "bin_count"], **kwargs):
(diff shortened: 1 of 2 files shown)


nslsii/areadetector/xspress3.py
```diff
@@ -136,11 +136,14 @@ class Xspress3ExternalFileReference(Signal):
 
     """
 
-    def __init__(self, *args, dtype_str="uint32", bin_count=4096, dim_name="bin_count", **kwargs):
+    def __init__(self, *args, dtype_str="uint32", bin_count=4096, dim_name=["frame_number", "number_channels", "bin_count"], **kwargs):
```
danielballan, line 139: We should avoid mutable defaults this can be a tuple.

nslsii/areadetector/xspress3.py
```diff
@@ -136,11 +136,14 @@ class Xspress3ExternalFileReference(Signal):
...
         super().__init__(*args, **kwargs)
         self.dtype_str = dtype_str
         self.shape = (bin_count,)
-        self.dims = (dim_name,)
+        if isinstance(dim_name, str):
+            self.dims = (dim_name,) # Keeping this option for backwards compatibility with some beamlines
+        else:
+            self.dims = (*dim_name,)
```
danielballan, line 146: If `dim_name` is a tuple to begin with, as suggested above, this can just be `self.dims = dim_name`.

danielballan (changes requested):

jmaruland:
@mrakitin Can you review and merge this for me, please?
I added Dan's reviews already.

nslsii/tests/test_xspress3.py
```diff
@@ -143,7 +143,7 @@ def test_instantiate_channel_class():
 
     assert channel_2.image.dtype_str == "uint32"
     assert channel_2.image.shape == (4096,)
-    assert channel_2.image.dims == ("bin_count",)
+    assert channel_2.image.dims == ('frame_number', 'number_channels', 'bin_count',)
```
mrakitin, line 146: `black` prefers double quotes. Can we switch those?
```suggestion
    assert channel_2.image.dims == ("frame_number", "number_channels", "bin_count",)
```

nslsii/areadetector/xspress3.py
```diff
@@ -136,11 +136,16 @@ class Xspress3ExternalFileReference(Signal):
...
     """
 
-    def __init__(self, *args, dtype_str="uint32", bin_count=4096, dim_name="bin_count", **kwargs):
+    def __init__(self, *args, dtype_str="uint32", bin_count=4096, dim_name=("frame_number", "number_channels", "bin_count"), **kwargs):
         super().__init__(*args, **kwargs)
         self.dtype_str = dtype_str
         self.shape = (bin_count,)
-        self.dims = (dim_name,)
+        if isinstance(dim_name, str):
+            self.dims = (dim_name,) # Keeping this option for backwards compatibility with some beamlines
```
mrakitin, line 144: Another style comment:
```suggestion
            self.dims = (dim_name,)  # Keeping this option for backward compatibility with some beamlines
```

nslsii/areadetector/xspress3.py
```diff
@@ -136,11 +136,16 @@ class Xspress3ExternalFileReference(Signal):
...
+        elif isinstance(dim_name, list):
+            self.dims = (*dim_name,)
```
mrakitin, line 146: Why not `tuple(dim_name)` here?

mrakitin (commented):
Looks good overall, just added a couple of non-blocking comments.

nslsii/areadetector/xspress3.py
jmaruland: In this case, the name is coming as a tuple already. We just pass it as it is to the next process

nslsii/areadetector/xspress3.py
jmaruland: Changed to tuple

PR #179, opened 2024-05-30
Add `__main__`

danielballan:
This enables:

```
$ python -m nslsii.sync_experiment --help
usage: __main__.py [-h] -b BEAMLINE -p PROPOSAL [-v | --verbose | --no-verbose]

Start or switch beamline experiment and record it in Redis

options:
  -h, --help            show this help message and exit
  -b BEAMLINE, --beamline BEAMLINE
                        Which beamline (e.g. CHX)
  -p PROPOSAL, --proposal PROPOSAL
                        Which proposal (e.g. 123456)
  -v, --verbose, --no-verbose
15:44 dallan@tensorbook:nslsii 
```

and also usage with `pdb` like so:

```
python -m pdb -m nslsii.sync_experiment --help
```

### DIFF
--- a/nslsii/sync_experiment/__main__.py
+++ b/nslsii/sync_experiment/__main__.py
@@ -1,0 +1,4 @@
+from .sync_experiment import main
+
+if __name__ == "__main__":
+    main()


mrakitin (approved):

jwlodek:
I would suggest also adding a version argument for both sync_experiment, and maybe for the package as a whole, like we did with `bluesky`.

mrakitin:
> I would suggest also adding a version argument for both sync_experiment, and maybe for the package as a whole, like we did with `bluesky`.

Decided to do it via a separate PR.

Issue #180, opened 2024-05-30
Add more proposal info


danielballan:
Include proposal title, number (i.e. without the `pass-` part), and PI last name.

PR #181, opened 2024-05-30
Added proposal info to redis

genematx:
Closes Issue #180. Added more info about proposals to the Redis dictionary as required by SRX.

### DIFF
--- a/nslsii/sync_experiment/sync_experiment.py
+++ b/nslsii/sync_experiment/sync_experiment.py
@@ -115,7 +115,7 @@ def sync_experiment(proposal_number, beamline, verbose=False, prefix=""):
         pi_name = ""
         for user in users:
             if user.get("is_pi"):
-                pi_name = f'{user.get("first_name", "")} {user.get("last_name", "")}'
+                pi_name = f'{user.get("first_name", "")} {user.get("last_name", "")}'.strip()
 
         md["data_session"] = new_data_session
         md["username"] = username
--- a/nslsii/sync_experiment/sync_experiment.py
+++ b/nslsii/sync_experiment/sync_experiment.py
@@ -112,10 +112,16 @@ def sync_experiment(proposal_number, beamline, verbose=False, prefix=""):
         if not should_they_be_here(username, new_data_session, beamline):
             raise AuthorizationError(f"User '{username}' is not allowed to take data on proposal {new_data_session}")
 
+        pi_name = ""
+        for user in users:
+            if user.get("is_pi"):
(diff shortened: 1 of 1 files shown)


nslsii/sync_experiment/sync_experiment.py
```diff
@@ -112,10 +112,16 @@ def sync_experiment(proposal_number, beamline, verbose=False, prefix=""):
...
+        pi_name = ""
+        for user in users:
+            if user.get("is_pi"):
+                pi_name = " ".join([user.get("first_name", ""), user.get("last_name", "")])
+
         md["data_session"] = new_data_session
         md["username"] = username
         md["start_datetime"] = datetime.now().isoformat()
         md["cycle"] = get_current_cycle()
+        md["proposal"] = {"number": proposal_number, "title": proposal_data.get("title"), "pi_name": pi_name}
```
genematx, line 124: A schema question... Should proposal number be a string (instead of an integer), and maybe called `id` (to match the API output)?

nslsii/sync_experiment/sync_experiment.py
```diff
@@ -112,10 +112,16 @@ def sync_experiment(proposal_number, beamline, verbose=False, prefix=""):
         if not should_they_be_here(username, new_data_session, beamline):
             raise AuthorizationError(f"User '{username}' is not allowed to take data on proposal {new_data_session}")
 
+        pi_name = ""
+        for user in users:
+            if user.get("is_pi"):
+                pi_name = user.get("first_name", "") + " " + user.get("last_name", "")
```
mrakitin, line 118: Why not f-string?
```suggestion
                pi_name = f'{user.get("first_name", "")} {user.get("last_name", "")}'
```

nslsii/sync_experiment/sync_experiment.py
mrakitin: It makes sense to me, in case the number is not a number in the future (such as a uid string or something else).

mrakitin (commented):
A small suggestion and a comment are below. Good otherwise.

nslsii/sync_experiment/sync_experiment.py
genematx: ah, yes of course! I was looking at this line and something just felt so wrong, but I couldn't see why...🤦‍♂️ thanks!

nslsii/sync_experiment/sync_experiment.py
genematx: Done. Now `['proposal']['id']` is a string.

mrakitin (approved):
Looks good to me now!

PR #182, opened 2024-05-31
Add a CLI tool to get IOC hostname for a given PV

mrakitin:
Originally implemented via https://github.com/NSLS-II-HEX/hextools/pull/19, moving to the facility-wide package.

### DIFF
--- a/nslsii/epics_utils.py
+++ b/nslsii/epics_utils.py
@@ -34,4 +34,4 @@ def main():
 
     args = parser.parse_args()
     hostname = get_ioc_hostname(args.pvname)
-    print(f"IOC hostname for {args.pvname} is {hostname}")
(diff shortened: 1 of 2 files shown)


nslsii/epics_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    except (gaierror, herror):
+        hostname = socket.gethostbyaddr(epics_addr)[0]
+
+    return hostname
+
+
+def main():
+    # Used by the `sync-experiment` command
+
+    parser = argparse.ArgumentParser(description="Start or switch beamline experiment and record it in Redis")
```
jwlodek, line 32: Looks like description needs to be updated?

jwlodek (approved):

PR #186, opened 2024-07-18
MNT: use packaging.version instead of distutils

tacaswell:


tacaswell:
Dropped the attempt to test on py312 because it requires getting bluesky-kafka to work with py312, blueksy-kafka explicitly uses intake in its `conftest.py` to set up the fixtures, and intake imports distutils.

mrakitin (approved):
Thanks for the fix, @tacaswell!

PR #187, opened 2024-07-19
Fix unbound variable in sync-experiment

thomaswmorris:
This should stop an unbound variable bug and still let the run start (with no proposal data and a warning).


genematx (approved):
Thanks for fixing this, Tom!

PR #188, opened 2024-07-24
Fix for Python 3.12

RobertSchaffer1:


RobertSchaffer1:
@mrakitin @tacaswell I'm not sure if this suffices, but if we import the setuptools module it carries with it a copy of the distutils module. This resolves the issue in the short term, in the longer term I think the easiest is to wait for the intake dependency to resolve the conflict.

PR #190, opened 2024-08-07
Enable CLI to bypass required flag "--beamline" if specified in env

### LINKED ISSUE #189: Use  `BEAMLINE_ACRONYM` as default
We now define `BEAMLINE_ACRONYM` in the workstation environment via [profile.d](https://github.com/NSLS2/ansible/blob/26e07dc52c0b87f616317f91f9a36fce40a48668/roles/profile/templates/beamline_acronym.sh.j2).

The `sync_experiment` module should be updated to use that as a default value, if set. https://github.com/NSLS-II/nslsii/blob/master/nslsii/sync_experiment/sync_experiment.py

Also, in the directory there is a stray `.DS_Store` file, accidentally committed, that should be cleaned up.

Kezzsim:
Resolves #189 by adding additional parsing logic to the CLI component which will get the beamline from the environment variable `BEAMLINE_ACRONYM`. 

Some design consciderations:
- This is implemented in the CLI args parser section as to allow the user to still override the env var with an argument if needed, this seems like it would be more intuitive.
- Because of this decision, when invoking the function from within another python program; beamline will still need to be provided. This shouldn't be a huge deal because python scripting is automated and will be run by the machine and not a human. This entire endeavor is an ergonomic concern.


nslsii/sync_experiment/sync_experiment.py
```diff
@@ -188,6 +188,16 @@ def main():
     parser.add_argument("-v", "--verbose", action=argparse.BooleanOptionalAction)
 
     args = parser.parse_args()
+    
+    # Read from env:
```
danielballan, line 192: One simplification:

```py
    # Read from env:
    beamline = os.getenv("BEAMLINE_ACRONYM")
    # Override if in command line args:
    if args.beamline:
        beamline = args.beamline
```

And another:
```py
# Commandline args take precedence over env
beamline = args.beamline or os.getenv("BEAMLINE_ACRONYM")
```

Kezzsim: Addressed in 4fdd2aa

PR #193, opened 2024-09-13
Fix sync_experiment for sst, enable custom prefix

nmaytan:


nslsii/sync_experiment/sync_experiment.py
```diff
@@ -177,6 +182,14 @@ def main():
         help="Which beamline (e.g. CHX)",
         required=True,
     )
+    parser.add_argument(
+        "-e",
```
padraic-shafer, line 186: Although `-p` Is already used, `-e` is non-intuitive. It's probably better to have no shortcut rather than a confusing one.
```suggestion
```

padraic-shafer (changes requested):
Overall looks good.  I recommend using `--prefix` flag without the `-e` shortcut.

nmaytan:
I figured folks would think `-e` for "endstation" when using the script. That said, I have no problem omitting the shorthand. @danielballan @padraic-shafer

nmaytan:
I think I agree with this. It is misleading slightly as `endstation` sort of implies that it should be a "known entity" as beamline is, but it is actually arbitrary. That said, in the context of sync-experiment, it should virtually always mean endstation, I think...

Along those lines, maybe one day we want to make sync-experiment match this option to the env `$ENDSTATION_ACRONYM`

prjemian:
Perhaps `REDIS_PREFIX` is more precise?  Unless you intend other use of this term.

nmaytan:
That might make sense, though I could imagine wanting to append the prefix in other ways besides just the Redis client. (such as for log entries).

Really treating this as "tell us your endstation" and not just "tell us some prefix for redis" has at least one other benefit- we could take care of the separator ourselves. I can very easily anticipate someone accidentally entering `mycoolendstation` instead of `mycoolendstation-`  (or just using an inconsistent/different separator) and effectively plopping into a different Redis namespace.

prjemian:
Last question for me is: Will you ever add additional complexity into the end station such that this prefix is not sufficient?  (Some of our end stations host more than one scientific instrument.  Some of them leave empty space for swapping in/out other major instruments.)

nmaytan:
Good question Pete (nice to meet you, by the way). I am perhaps not the ultimate authority on this, but I expect we will only be separating this metadata by endstation and not per each instrument at the endstation. Certainly we'd be interested if you have any thoughts or concerns.

danielballan (approved):

nslsii/sync_experiment/sync_experiment.py
```diff
@@ -111,9 +111,15 @@ class AuthorizationError(Exception): ...
...
-    redis_client = redis.Redis(host=f"info.{beamline.lower()}.nsls2.bnl.gov")
+    normalized_beamlines = {
+        "sst1": "sst",
+        "sst2": "sst",
+    }
+    redis_beamline = normalized_beamlines.get(beamline.lower(), beamline)
+    redis_client = redis.Redis(host=f"info.{redis_beamline.lower()}.nsls2.bnl.gov")
 
-    md = RedisJSONDict(redis_client=redis_client, prefix=prefix)
+    redis_prefix = f"{prefix}-" if prefix else ""
```
genematx, line 121: Is there a reason we add "-" to the prefix here? If this is a redis's syntectic requirement, would it be better to handle this by `RedisJSONDict` itself? Right now, to run `sync-experiment`, one uses `-e swaxs` (without the '-'), but in the profile collection, we have to set `prefix='swaxs-'` when we initialize the redis client, which is confusing and error-prone.

nmaytan: I don't have a strong opinion, particularly if implemented this way (i.e. it is optional to use) I don't think it can hurt.

genematx: I think specifying a delimiter with a default value set to "-" makes a lot of sense!

PR #196, opened 2024-09-27
Make sync-experiment work for commissioning proposals

nmaytan:


nslsii/sync_experiment/sync_experiment.py
```diff
@@ -24,6 +24,17 @@ def get_current_cycle() -> str:
     return cycle_response.json()["cycle"]
 
 
+def get_commissioning_status(proposal_number, beamline) -> bool:
+    commissioning_proposals_response = nslsii_api_client.get(
```
padraic-shafer, line 28: As this project continues to grow, it will be helpful to add docstrings, even to small functions.

```suggestion
def get_commissioning_status(proposal_number, beamline) -> bool:
    """True if proposal_number is registered as a commissioning proposal; else False."""
    commissioning_proposals_response = nslsii_api_client.get(
```

padraic-shafer (approved):
Aside from adding docstrings, my only comment is that the name `get_commissioning_status()` implies that some sort of status object will be returned.  Something like `is_commissioning_proposal()` would more directly convey that it returns a boolean based on membership.

nmaytan:
Thanks @padraic-shafer, addressed these comments

padraic-shafer (approved):

PR #198, opened 2024-10-04
Refactor auth mechanism in sync-experiment

genematx:
Context: Some beamlines (e.g. SMI) collect data on multiple proposals during the same run to make use of the same environment conditions between the experiments (e.g. temperature); changing these conditions often would unnecessarily require too much downtime for the instruments.
Our current `sync-experiment` workflow assumes that a user (or a beamline scientist) switches to a new proposal at the beginning of acquisition (possibly even before opening `bsui`). This would be impractical for highly dynamic or automated plans as above.

This PR Separates the authentication mechanism in the internal `sync_experiment` function from the updating of the redis dictionary. This would allow us to switch between multiple proposals within a profile_collection plan (using the new `switch_redis_proposal(proposal_number, beamline, username, prefix)` function -- SMI already have [something very similar](https://github.com/NSLS-II-SMI/profile_collection/blob/8edf9752102dc37e52b63d483e3722aac7123700/startup/80-config.py#L52) that they extensively use in their profile collection, and the idea is to just replace it). This function will still check if a user has the rights to collect data on a proposal, but wouldn't ask for their password. We will need that before running `bsui` or in the base step of profile collection, the user properly authenticates with `sync-experiment` (this is effectively an honor system, but writing the data in a correct location is is in their interests, because otherwise they won't be able to read it).

These refactoring changes will not break any existing deployments.

### DIFF
--- a/nslsii/sync_experiment/sync_experiment.py
+++ b/nslsii/sync_experiment/sync_experiment.py
@@ -124,7 +124,7 @@ def should_they_be_here(username, new_data_session, beamline):
 class AuthorizationError(Exception): ...
 
 
-def sync_experiment(proposal_number, beamline, verbose=False, prefix=""):
+def sync_experiment(proposal_number, beamline, username, verbose=False, prefix=""):
 
     normalized_beamlines = {
         "sst1": "sst",
@@ -137,28 +137,23 @@ def sync_experiment(proposal_number, beamline, verbose=False, prefix=""):
     md = RedisJSONDict(redis_client=redis_client, prefix=redis_prefix)
 
     new_data_session = f"pass-{proposal_number}"
-    username = input("Username : ")
 
     if (new_data_session == md.get("data_session")) and (
         username == md.get("username")
     ):
-
         warnings.warn(
             f"Experiment {new_data_session} was already started by the same user."
         )
 
     else:
 
-        proposal_data = validate_proposal(new_data_session, beamline)
-        users = proposal_data.pop("users")
-
-        authenticate(username)
-
         if not should_they_be_here(username, new_data_session, beamline):
             raise AuthorizationError(
                 f"User '{username}' is not allowed to take data on proposal {new_data_session}"
             )
 
+        proposal_data = validate_proposal(new_data_session, beamline)
+        users = proposal_data.pop("users")
         pi_name = ""
         for user in users:
             if user.get("is_pi"):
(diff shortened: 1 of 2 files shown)


nslsii/sync_experiment/sync_experiment.py
```diff
@@ -124,41 +125,53 @@ def should_they_be_here(username, new_data_session, beamline):
...
-
+def switch_redis_proposal(
+    proposal_number: Union[int, str],
+    beamline: str,
+    username: Optional[str] = None,
+    prefix: str = "",
+) -> RedisJSONDict:
+    """Update information in RedisJSONDict for a specific beamline
+
+    Paraneters
```
mrakitin, line 136:
```suggestion
    Parameters
```

nslsii/sync_experiment/sync_experiment.py
```diff
@@ -124,41 +125,53 @@ def should_they_be_here(username, new_data_session, beamline):
...
+    ----------
+    proposal_number : int or str
+        number of the desired proposal, e.g. `123456`
+    beamline : str
+        normalized beamline acronym, case-insensitive, e.g. `SMI` or `sst`
+    username : str or None
+        login name of the user assigned to the proposal; if None, current user will be kept
+    prefix : str
+        optional prefix to identigy a specific endstation, e.g. `opls`
```
mrakitin, line 145:
```suggestion
        optional prefix to identify a specific endstation, e.g. `opls`
```

mrakitin (dismissed):
Looks good to me. Let's correct a couple of typos below.

mrakitin:
@danielballan, what are your thoughts on that? @nmaytan expressed some concerns about the approach.

genematx:
Thanks for bumping this, @mrakitin! Would be great to resolve this before next Monday when we come back to SMI. Or perhaps think of another strategy of how we are going to handle their concurrent experiments from multiple proposals (queueserver, maybe?).

nslsii/sync_experiment/sync_experiment.py
```diff
@@ -204,11 +219,31 @@ def sync_experiment(proposal_number, beamline, verbose=False, prefix=""):
...
+        "sst1": "sst",
+        "sst2": "sst",
+    }
+    redis_beamline = normalized_beamlines.get(beamline.lower(), beamline)
+
+    md = switch_redis_proposal(
+        proposal_number, beamline=redis_beamline, username=username, prefix=prefix
+    )
+
+    if verbose:
```
genematx, line 243: Do we still need this?

danielballan (approved):

danielballan:
Moving ahead to un-block things. We will be iterating on this in the near future.

PR #199, opened 2024-10-29
Update authentication method for sync experiment to be more robust in…

jwlodek:
… the case of dc server being down

### DIFF
--- a/nslsii/sync_experiment/sync_experiment.py
+++ b/nslsii/sync_experiment/sync_experiment.py
@@ -1,5 +1,5 @@
 from ldap3 import Server, Connection, NTLM
-from ldap3.core.exceptions import LDAPInvalidCredentialsResult
+from ldap3.core.exceptions import LDAPInvalidCredentialsResult, LDAPSocketOpenError
 
 import json
 import re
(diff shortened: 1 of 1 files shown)


nslsii/sync_experiment/sync_experiment.py
```diff
@@ -88,21 +88,31 @@ def validate_proposal(data_session_value, beamline) -> Dict[str, Any]:
...
+    for server in ["1", "2", "3"]:
+        if authenticated:
+            break
+        auth_server = Server(f"dc{server}.bnl.gov", use_ssl=True)
+
+        try:
+            connection = Connection(
+                auth_server,
+                user=f"BNL\\{username}",
+                password=getpass("Password : "),
```
genematx, line 101: Wouldn't it ask for the password (at most) three times?

genematx: lol, just noticed it's merged already :)

jwlodek: In the case of unable to connect, it never gets to the password prompt, just fails. So you should just get one password prompt no matter what

PR #203, opened 2024-10-29
CI: only use the `published` event for PyPI releases

mrakitin:
This PR is to follow up on #197 (per https://stackoverflow.com/a/61066906) to avoid a duplicate execution of the same CD workflow, as happened with the v0.10.6 release:

image


jwlodek (approved):

padraic-shafer (approved):
Let's try it.  Results in GH issue discussions were mixed, but this seems like a reasonable  option to me.

PR #205, opened 2024-10-30
fix: default broker to None

maffettone:
Data security changes are creating a subscription to a `db.insert` and then promptly removing this in favor a tiled writing client. See [BMM Profile](https://github.com/NSLS-II-BMM/profile_collection/blob/e4f3670f5de7d6d91c9e26c89e2a66197a2670a8/startup/BMM/user_ns/base.py#L41-L42). 

This is not only sloppy, but causes failures in any services that cannot create a broker in that fashion (e.g., QueueServer).

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -28,7 +28,7 @@ def import_star(module, ns):
 
 def configure_base(
     user_ns,
-    broker_name,
+    broker_name=None,
     *,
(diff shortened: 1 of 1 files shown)


jwlodek (approved):

mrakitin (approved):
I like the approach much better. Thanks for the fix, @maffettone!

PR #208, opened 2024-11-01
fix configure_kafka_publisher call

### LINKED ISSUE #207: configure_kafka_publisher not working when Broker instance passed to configure_base
`broker_name` can be a `Broker` instance, or a string in the call to `configure_base`, but `configure_kafka_publisher` expects a string. So `configure_kafka_publisher` will fail if you pass a` Broker` to `configure_base`.

https://github.com/NSLS-II/nslsii/blob/8a952c23c07bb0e4eeec0c075d74bd60bf09f38f/nslsii/__init__.py#L74

gwbischof:
Fixes #207

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -223,7 +223,10 @@ def configure_base(
         configure_ipython_logging(exception_logger=log_exception, ipython=ipython)
 
     if publish_documents_with_kafka:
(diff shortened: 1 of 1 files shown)


jwlodek (approved):

mrakitin:
Closed via #209.

PR #209, opened 2024-11-06
fix: configure_kafka_publisher assumed a string

maffettone:
`configure_kafka_publisher` assumed that `broker_name` was a string. This broke silently at BMM when using RE as a service without a broker configured, but kafka set to True.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -96,9 +96,10 @@ def configure_base(
     ipython_logging : boolean, optional
         True by default. Console output and exception stack traces will be
         written to IPython log file when IPython logging is enabled.
-    publish_documents_with_kafka: boolean, optional
+    publish_documents_with_kafka: Union[boolean, str], optional
         False by default. If True publish bluesky documents to a Kafka message broker using
-        configuration parameters read from a file.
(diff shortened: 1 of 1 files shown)


maffettone:
tagging @gwbischof for visibility, since I didn't realize he'd already attacked this elsewhere.

nslsii/__init__.py
```diff
@@ -223,7 +224,15 @@ def configure_base(
         configure_ipython_logging(exception_logger=log_exception, ipython=ipython)
 
     if publish_documents_with_kafka:
-        configure_kafka_publisher(RE, beamline_name=broker_name)
+        if isinstance(publish_documents_with_kafka, str):
+            configure_kafka_publisher(RE, beamline_name=publish_documents_with_kafka)
+        elif hasattr(broker_name, "name"):
+            configure_kafka_publisher(RE, beamline_name=broker_name.name)
```
mrakitin, line 230: In #208, @gwbischof proposed to `.lower()` the name just in case it's not lower-case for some reason.
```suggestion
            configure_kafka_publisher(RE, beamline_name=broker_name.name.lower())
```

mrakitin (commented):
All looks good - just spotted a minor difference between the solution in #208.

mrakitin (approved):

PR #211, opened 2025-03-12
Adds if_touch_beamline function to common

jennmald:
Adds the `if_touch_beamline` function from SRX profile collection to the common folder. This will help with the efforts of https://github.com/NSLS2/test-beamline-profiles


nslsii/common/ipynb/touchbl.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+def if_touch_beamline(envvar="TOUCHBEAMLINE"):
+    value = os.environ.get(envvar, "false").lower()
+    if value in ("", "n", "no", "f", "false", "off", "0"):
+        return False
+    elif value in ("y", "yes", "t", "true", "on", "1"):
+        return True
+    else:
+        raise ValueError(f"Unknown value: {value}")
```
mrakitin, line 10: You may need to add a new line here and in the next file.

nslsii/tests/test_ipynb.py
```diff
@@ -6,3 +6,8 @@ def test_ipynb():
     obj.metadata
     obj.url
     obj.reload()
+
+def test_touchbl():
+    from nslsii.common.ipynb import if_touch_beamline
+    if if_touch_beamline():
+        pass
```
mrakitin: Let's use `assert` to test that if the `TOUCHBEAMLINE` env var is not set, the function returns `False`, otherwise - `True`. The env var can be set via `os.environ["TOUCHBEAMLINE"] = ...` for testing inside of the test. We can test multiple inputs as well via `@pytest.mark.parametrize`.

nslsii/tests/test_ipynb.py
```diff
@@ -6,3 +6,8 @@ def test_ipynb():
     obj.metadata
     obj.url
     obj.reload()
+
+def test_touchbl():
+    from nslsii.common import if_touch_beamline
+    if if_touch_beamline():
+        pass
```
mrakitin: Let's use assert to test that if the `TOUCHBEAMLINE` env var is not set, the function returns `False`, otherwise - `True`. The env var can be set via `os.environ["TOUCHBEAMLINE"] = ...` for testing inside of the test. We can test multiple inputs as well via `@pytest.mark.parametrize`.

mrakitin (commented):
You may also consider adding the relative import to `nslsii/common/__init__.py` to make it easier to import the function.

mrakitin:
Power-cycled the PR to start the CI jobs after re-enabling GHA in this repo.

nslsii/tests/test_ipynb.py
```diff
@@ -6,3 +6,9 @@ def test_ipynb():
     obj.metadata
     obj.url
     obj.reload()
+
+def test_touchbl():
+    from nslsii.common import if_touch_beamline
```
mrakitin, line 11: To enable this import, we need to add a line like that in `nsls2/common/__init__.py`:
```py
from .touchbl import if_touch_beamline
```

PR #212, opened 2025-03-14
Remove persistent dict and add redis_json_dict support

jennmald:
Stops the tests from failing because of `zict` dependency plus upgrades from `PersistentDict` to `redis_json_dict`


jwlodek (changes requested):

mrakitin (approved):
Looks good to me!

PR #214, opened 2025-03-27
Move redis parameters to the kwarg-only section

### LINKED ISSUE #213: Fix `configure_base` for backward compatibility
I am seeing problems with CI using the new version of the package: https://github.com/NSLS2/nsls2-collection-tiled/actions/runs/14098960313/job/39491728033?pr=53.

We need to move the new redis-related parameters to the kwarg-only section, so that we can still use the positional argument for the broker name, such as in the CSX profile:
```py
nslsii.configure_base(ip.user_ns, 'csx', publish_documents_with_kafka=True, bec=False)
```

jennmald:
Moves the redis parameters to the kwarg-only section to resolve #213

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -26,8 +26,6 @@ def import_star(module, ns):
 
 def configure_base(
     user_ns,
-    redis_url=None,
-    redis_prefix="",
     broker_name=None,
     *,
     bec=True,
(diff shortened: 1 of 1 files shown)


mrakitin (approved):
Thank you for the fix, @jennmald!

PR #218, opened 2025-04-14
Add helpful re debug subs

jwlodek:


nslsii/re_subs.py
```diff
@@ -0,0 +1,65 @@
...
+            self.output_file = None
+
+
+class BlueskyDocStreamPrinter:
+    """
+    Print documents to stdout.
+    """
+
+    def __init__(self):
+        self.print_docs_to_stdout = False
```
genematx, line 58: Do I understand this correctly, is this parameter meant to be set externally after the class initialization? Just wondering, what's the purpose of it?

nslsii/re_subs.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, write_directory: Path):
+        if not os.path.exists(write_directory):
+            raise FileNotFoundError(
+                f"Directory does not exist: {write_directory}"
+            )
+        elif not os.access(write_directory, os.W_OK):
+            raise PermissionError(
+                f"Cannot write to directory: {write_directory}"
+            )
+        self.write_json_file: bool = False
```
genematx, line 17: same as below

nslsii/re_subs.py
```diff
@@ -0,0 +1,65 @@
...
+        self.write_directory = write_directory
+        self.output_file: TextIOWrapper | None  = None
+        self.document_cache = []
+
+    def __call__(self, name: str, doc: dict):
+        if self.write_json_file:
+            if name == "start":
+                self.output_file = open(
```
genematx, line 25: I wonder if there's a way to refactor it using a context manager instead of keeping the file open.

genematx (approved):
Looking good! thanks, Jakub for this. I think there was an idea before to have `JSONWriter` in bluesky callbacks, but I've been to lazy to put it there. Do you think, it would still be useful to have a callback? Could it be reused here (I'm not that familiar with nslsii subs).
On a separate note, as Dan mentioned before, [JSON Lines](https://jsonlines.org/) format could be particularly useful for storing the messages (jsonl is meant to be appendable). In my workflows, I personally still prefer the usual JSON because it can be easily formatted with VSCode (at least a year ago, it didn't understand jsonl), but JSONL should definitely be more efficient and might be more suitable for debugging.

nslsii/re_subs.py
jwlodek: Good idea, I can do that for sure.

jennmald (approved):
It looks good, I like Eugene's idea of doing the refactor. I tried to fix the CI but was unsuccessful.

jwlodek:
CI is broken due to an issue w/ the most recent release of ophyd - Tom's PR needs to be merged + new release made to fix it.

jennmald (approved):

nslsii/re_subs.py
genematx: I think the toggle-able methods make the behavior more obvious, especially if the user is expected to do this dynamically at runtime.

The dedicated method `set_output_directory` might be a bit confusing because it adds another private attribute not directly visible from `__init__`, but then it makes the init cleaner, so pros and cons...
Could be something like `self._write_directory = self.set_output_directory(write_directory: str)` if it returned it back, to make it extra explicit. Also it might better to call it `set_write_directory` for consistency, but that's probably nitpicking.

nslsii/re_subs.py
```diff
@@ -0,0 +1,96 @@
...
+class BlueskyDocJSONWriter:
+
+    def __init__(self, write_directory: Path, flush_on_each_doc: bool = True):
+        self.set_output_directory(write_directory)
+        self._write_json_file: bool = False
+        self._flush_on_each_doc = flush_on_each_doc
+        self._output_file_name: str | None  = None
+        self._document_cache = []
+
+    def set_output_directory(self, write_directory: Path):
```
genematx, line 15: The dedicated method `set_output_directory` might be a bit confusing because it adds another private attribute not directly visible from `__init__`, but then it makes the init cleaner, so pros and cons...
Could be something like `self._write_directory = self.set_output_directory(write_directory: str)` if it returned it back, to make it extra explicit. Also it might better to call it `set_write_directory` for consistency, but that's probably nitpicking.

genematx (approved):

PR #227, opened 2025-05-16
add option for broker_name to be a TiledWriter object

### LINKED ISSUE #226: Update configure base to not expect databroker insert
Currently, there is a way to have a Tiled inserter that allows `configure_base` to not fail. This requires creating a new `TiledInserter` class to be passed where `broker_name` is in the args of `configure_base`.

We need to:

- [ ] update configure base to not expect the `.insert` function
- [ ] Type `broker_name` with string and whatever client it expects

jennmald:
Attempting to resolve #226 
I figured the best way to stay consistent with backwards compatibility would be to keep `db` defined the same way but later check if it is an object of TiledWriter. The subscribe directly to that instead of calling `.insert`. This should alleviate the problem of creating the `TiledInserter` during data security.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -164,12 +164,11 @@ def configure_base(
         db = broker_name
 
     if db is not None:
-        from bluesky.callbacks.tiled_writer import TiledWriter
-        if isinstance(db, TiledWriter):
-            # subscribe to the TiledWriter
-            RE.subscribe(db)
-        else:
+        if isinstance(db, Broker):
+            # subscribe to the Broker
             RE.subscribe(db.insert)
+        else:
+            RE.subscribe(db)
 
     if pbar:
         # Add a progress bar.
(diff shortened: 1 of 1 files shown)


danielballan:
This is great, thank you! Compare the consequences of checking for an instance of the old thing versus checking for an instance of the new thing.

During the Python 2 to 3 transition, we were encouraged to check for 2 rather than check for 3. The reason was, we do not want code to revert to the old behavior when Python 4 comes out.

jennmald:
> This is great, thank you! Compare the consequences of checking for an instance of the old thing versus checking for an instance of the new thing.
> 
> During the Python 2 to 3 transition, we were encouraged to check for 2 rather than check for 3. The reason was, we do not want code to revert to the old behavior when Python 4 comes out.

This makes a lot of sense, thank you @danielballan ! I will think about this and update the PR.

danielballan (approved):

tacaswell:
Merging over kafka test failures.

PR #231, opened 2025-09-17
Allow for passing through windows drive letter to NSLS2PathProvider

jwlodek:
Updates the `NSLS2PathProvider` to support passing in a separate write directory. Also simplifies the logic for the base directory setup. Removes the `AcqModeFilenameProvider`, replaced instead by the `REMetadataFilenameProvider` that allows for creating filenames from formatted RE metadata.


nslsii/ophyd_async/providers.py
```diff
@@ -157,31 +190,25 @@ class NSLS2PathProvider(PathProvider):
...
-        ) + (beamline_tla_suffix or "")
-
-        self._beamline_proposals_dir = (
-            Path("/nsls2/data/") / beamline_data_dir / "proposals"
+        self._granularity = granularity if isinstance(granularity, YMDGranularity) else YMDGranularity[granularity]
+        self._base_read_directory = self._get_beamline_proposals_dir(beamline_data_dirname=beamline_data_dirname)
+        self._base_write_directory = (
+            self._base_read_directory
+            if not windows_drive_letter
+            else _drive_letter_to_path(windows_drive_letter) / "proposals"
```
thopkins32: This assumes that a drive letter corresponds to a specific path, but is this standard? I think I've mounted all the way up to `<path>/proposals` for the drive letter to prevent other not relevant paths from being mounted onto the Windows IOC.

thopkins32 (approved):
LGTM. Just want to check one assumption about Windows drive letter and how it maps to a directory in practice.

nslsii/ophyd_async/providers.py
jwlodek: Would it make sense to be less magic and instead of taking in a drive letter, take in a `PureWindowsPath | None`?

thopkins32: I think that might be better. Let us pass the base path(s) as arguments.

thopkins32 (approved):

PR #234, opened 2025-11-18
Make sync-experiment Tiled-based

nmaytan:
Summary of changes so far:

- Keys in Redis are now always prefixed, even if there is no endstation ("fixes" greedy key ingestion that happens otherwise)
- The beamline and endstation acronyms are taken from the system environment, so that they don't have to be provided (but are overridable, as this won't work in some cases such as SST/SST1/SST2)
- Added facility parameter throughout to set us up for future LBMS support (still needs other work to actually function)
---
- Multiple proposals can now be activated at the same time, to support cases like at SMI where they interleave experiments. This is put in a new key `md['data_sessions_authorized']`
- Now that multiple proposals can be activated, we can enable switching proposals in a way that doesn't bypass auth as extremely -> `switch_proposal` let's you rotate through any proposals that are currently activated
- Adds the ability to switch between authorized proposals from the CLI
---
- Auth is totally Tiled-based, no more LDAP calls. This is done using `password_grant` only for now, but could use any auth flow that Tiled supports (sets us up for the future, i.e. OIDC)
- An API key is generated when `sync-experiment` runs, reducing the need to have multiple logins (API key will be used for subsequent Tiled clients)
- ~API keys are cached in Redis and managed with a priority queue based on expiration date. Keys are set to expire after 7 days.~
- ~5 keys can be cached at a time - each combination of username + proposal list counts as a new key~
- ~Only one API key is left unencrypted at a time, and it is put in Redis under a specific key so that it can be retrieved from other applications without decrypting it~
- ~The rest of the cached keys remain encrypted in Redis until they are needed~
- UPDATE: caching of API keys has been removed. Now, a single API key is stored in Redis unencrypted.
- Syncing or unsyncing the experiment now revokes the currently active API key (if one is found). If syncing, it is replaced with a new key.
- Generated keys are set to expire after 7 days.
---
- Adds the ability to "unsync" (logout) of an experiment


nslsii/sync_experiment/sync_experiment.py
```diff
@@ -1,292 +1,656 @@
...
+        dest="proposals",
+        nargs="+",
         type=int,
-        help="Which proposal (e.g. 123456)",
+        help="The proposal ID for the experiment",
         required=True,
     )
+    parser.add_argument(
+        "-s",
+        "--select",
```
genematx: should this be "activate", maybe?

nslsii/sync_experiment/sync_experiment.py
```diff
@@ -1,292 +1,656 @@
...
+            "or set the 'BEAMLINE_ACRONYM' environment variable."
+        )
+    for proposal_id in proposal_ids:
+        if not re.fullmatch(r"^\d{6}$", proposal_id):
+            raise ValueError(
+                f"Provided proposal ID '{proposal_id}' is not valid.\n "
+                f"A proposal ID must be a 6 character integer."
+            )
+    if select_proposal not in proposal_ids:
+        raise ValueError("Cannot select a proposal which is not being activated.")
```
genematx: I think we might better stick to either "select" or "activate" and use it consistently to avoid confusion. (same as comment below)
Or are they different concepts, actually?

nmaytan: Thanks for this feedback Eugene. As we discussed at the beamline, I agree with you that better words would be:
activated -> authenticated (you "authenticate" proposals for the beamline)
selected -> activated (you "activate" a specific proposal for an experiment)

I'll include this in the next round of refinement.

nslsii/sync_experiment/sync_experiment.py
```diff
@@ -1,292 +1,800 @@
...
+    if it is still fresh.
+    Failure to decrypt (e.g. a changed password) will also cause
+    existing keys to be ignored.
+
+    """
+    data_sessions_sanitized = sorted(
+        data_session.replace("-", "") for data_session in data_sessions
+    )
+    redis_prefix = f"{beamline}-{endstation}" if endstation else f"{beamline}"
+    redis_prefix = (
```
genematx: Does this mean that if a user authenticates to `pass-1` and `pass-2`, they get one api key, but later if they authenticate only to `pass-1` (or any other variation, really) a new key would need to be generated? Although, I see how checking subsets of proposals can quickly become complicated, but I wonder if there's a better way to track this.

nslsii/sync_experiment/sync_experiment.py
```diff
@@ -1,292 +1,800 @@
...
+            "No beamline provided! Please provide a beamline argument, "
+            "or set the 'BEAMLINE_ACRONYM' environment variable."
+        )
 
-    if server is None:
-        raise RuntimeError("Server name not found!")
+    beamline = beamline.lower()
+    if endstation:
+        endstation = endstation.lower()
+    normalized_beamline = normalized_beamlines.get(beamline.lower(), beamline.lower())
```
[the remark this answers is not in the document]
nmaytan: Thanks! Small oversight on my part of this redundancy. I did the same in other places that the normalization happens, so I will fix this in multiple places.

nslsii/sync_experiment/sync_experiment.py
```diff
@@ -1,292 +1,800 @@
...
+from datetime import datetime, timezone
+from getpass import getpass
+from pydantic.types import SecretStr
 from redis_json_dict import RedisJSONDict
+from tiled.client.context import Context, password_grant, identity_provider_input
+from tiled.profiles import load_profiles
+
+
+facility_api_client = httpx.Client(base_url="https://api.nsls2.bnl.gov")
+normalized_beamlines = {
```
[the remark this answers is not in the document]
nmaytan: Interesting idea, I agree that'd be nice. I'll give it some thought. I think I would lean towards some implementation of defaultdict, just to be "aggressive" that this should be a simple remapping of acronyms.

nslsii/sync_experiment/sync_experiment.py
```diff
@@ -1,292 +1,800 @@
...
+    Retrieve an API key from Redis.
+    Query Redis for key information, decrypt the API key if found,
+    ignore the key if it is expired, and finally return the key
+    if it is still fresh.
+    Failure to decrypt (e.g. a changed password) will also cause
+    existing keys to be ignored.
+
+    """
+    data_sessions_sanitized = sorted(
+        data_session.replace("-", "") for data_session in data_sessions
```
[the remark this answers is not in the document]
nmaytan: While it is true that for us, a data session = `pass-` + proposal number, this is not strictly true. Data sessions could in theory have different identifiers (even possibly disconnected from the literal proposal number). I made this specific to the data session in the spirit of that possibility.

nmaytan:
Updated to greedily revoke the existing API key rather than cache+rotate, enabled by the `revoke:apikeys` scope we added in Tiled for this purpose. This simplifies the tool quite a bit (no encryption needed and no priority queue in Redis).

This also gets us closer to ready for OIDC, where we wouldn't have had a password to encrypt the keys with. Probably just two more things needed?
* remove the `internal` mode check when creating the context
* move the username/password prompt out of the way

Tested current changes and seems to work well. More to come for OIDC when available to test with.

edit: also need to change root_node tag -> beamline tag when then change goes through server-side

thopkins32:
@nmaytan Can you try to merge this soon? We need ESM arpes to be on a new release of nslsii

nmaytan:
The generated API key is now scoped to include the `public` tag, instead of the `_ROOT_NODE_TLA` tag. The latter has been removed from Tiled's config, and the former is used to allow access into the beamline's top level catalog.

This may need some changing when we eventually allow users to create scans in the catalog as themselves.

PR #247, opened 2026-03-27
Greatly simplify the ophyd-async providers

jwlodek:


nslsii/ophyd_async/providers.py
```diff
@@ -1,61 +1,224 @@
...
         """
 
-        beamline_tla = (
-            os.getenv("ENDSTATION_ACRONYM", os.getenv("BEAMLINE_ACRONYM", "")).lower()
-            + self._tla_suffix
-        )
+        beamline_tla = os.getenv(
+            "ENDSTATION_ACRONYM", os.getenv("BEAMLINE_ACRONYM", "")
+        ).lower() + (self._tla_suffix or "")
         beamline_proposals_dir = Path(f"/nsls2/data/{beamline_tla}/proposals/")
```
jennmald: I'm not sure if this will cause the same error that I noticed at OPLS were the TLA is "PLS" instead of "SMI". Was this tested for beamlines with 2 end stations?

jennmald: @RobertSchaffer1 did we put in the permanent fix for this in sync-experiment?

jwlodek: I've modified the class to allow for passing in the TLA manaully, that should allow for working around these edge cases.

jennmald (approved):

jennmald (approved):

PR #249, opened 2026-04-08
Added function that searches for proposals by PI name

jmaruland:
This PR adds a function that will allow BL staff to search in tiled for proposals in by PI name and cycle (optional). The goal is to allow BL staff to import this method in their bsui environment and have this function at hand during their experiments.
This function was tested with the OPLS and SMI catalogs

### DIFF
--- a/nslsii/proposals.py
+++ b/nslsii/proposals.py
@@ -42,7 +42,7 @@ def find_proposals(client, pi_name, cycle=None, show_title=True):
 
     Example
     --------
-    >>> find_proposals(tiled_reading_client, 'Smith', cycle='2023-2')
+    >>> find_proposals(tiled_reading_client, 'Smith', cycle='2026-1')
     """
 
     results = client.search(Regex("proposal.pi_name", f"^{pi_name}"))
--- a/nslsii/proposals.py
+++ b/nslsii/proposals.py
@@ -51,35 +51,36 @@ def find_proposals(client, pi_name, cycle=None, show_title=True):
     proposal_distinct = results.distinct("proposal.proposal_id", counts=True)
 
     proposal_info = {}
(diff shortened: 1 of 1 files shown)


jmaruland:
After receiving some feedback from BL staff, I added an additional input parameter that allows optional queries. This is intended for other filed in the start document that BL staff would find useful to search for proposals. I narrowed down the search to string and numerical types; other data types are not included.
E.g.: 
```
>>> find_proposals(
        tiled_reading_client,
        "Smith",
        cycle="2026-1",
        optional_queries={"exposure_time": (">", 10)},
        show_title=True,
    )
```

mrakitin (approved):
It's a useful feature.

jwlodek:
Could we add some unit tests that cover this functionality? Also type hints for the function arguments would be nice

danielballan:
@jmaruland let me know that some queries time out (against SQL-backed catalogs). Better indexes are needed to support this. We aim to add those in the next maintenance (June 16) and it does not need to be a blocker for merging this PR.

I agree with @jwlodek that it's worth adding unit tests and type hints.

jmaruland:
Thanks for all the comments to improve this PR. I will add all these features next.

@danielballan my tests using the sql-backed catalogs have not passed this line that  involves the Like() method and unfortunately, this is the only line that I cannot test with the mongo-backed catalogs.
https://github.com/jmaruland/nslsii/blob/80b0b2e2ecfd1d66101baeaaf64f4024cab05eb6/nslsii/proposals.py#L69

dihenriksen (approved):

jmaruland:
@dihenriksen I just saw that you merged this PR a week ago. This PR was waiting for some work that @danielballan was doing with the sql database. Did you talk to him before merging this PR?
I would like to merge this branch, too. But without the indexing work in sql database, this PR has not been fully tested.

PR #250, opened 2026-04-15
ENH: if redis does not connect, suggest using ssl flag

tacaswell:


nslsii/sync_experiment/sync_experiment.py
```diff
@@ -179,17 +196,24 @@ def switch_redis_proposal(
...
-    location = prefix if prefix else redis_beamline
-    if redis_ssl:
-        redis_client = open_redis_client(redis_ssl=redis_ssl, redis_prefix=location)
-    else:
-        redis_url=f"info.{redis_beamline}.nsls2.bnl.gov"
-        redis_client = open_redis_client(redis_ssl=redis_ssl, redis_prefix=location, redis_url=redis_url)
+    redis_client = _get_client(beamline, redis_ssl, prefix)
+
+    try:
+        redis_client.ping()
```
mrakitin, line 202: If we use the secure server specified via the `REDIS_HOST` env var without the `-s` option, it hangs on this step indefinitely. We may need some sort of timeout.

mrakitin (changes requested):
Need to test it with an active proposal and multiple combinations of parameters. See the problem identified below.

mrakitin:
We also need to resolve the conflicts here.

PR #253, opened 2026-04-23
Disambiguate dual use of Redis prefix; Add redis_db option

thopkins32:
Redis prefix was being used for two purposes:
- To determine the location (endstation) to connect to the secure Redis instance
- As the prefix for filtering keys in `RedisJSONDict`

We are abandoning the use of filtering keys in `RedisJSONDict` in favor of using a separate Redis DB for multiple use-cases. Each endstation already has a dedicated secure Redis instance. Therefore, specifying the `endstation` in sync-experiment will resolve to the secure Redis server location. In `configure_base` you can now specify a Redis DB index to use. Finally, using the secure Redis instance (ssl enabled) and passing a prefix are mutually exclusive arguments which will raise a `ValueError` if attempted.

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -44,6 +44,7 @@ def configure_base(
     redis_port=6379,
     redis_ssl=False,
     redis_prefix="",
+    redis_db: int = 0,
 ):
     """
     Perform base setup and instantiation of important objects.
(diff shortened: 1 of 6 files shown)


RobertSchaffer1 (approved):
I didn't look at the tests too closely, but the rest looks good.

PR #254, opened 2026-05-05
FIX: use the correct way of specifying numpy dtypes

tacaswell:
Found by @padraic-shafer when trying to consume IOS data.


padraic-shafer (approved):
Thanks @tacaswell for fixing this.

PR #255, opened 2026-05-06
Fix open_redis_client auto deduction of redis name to use BEAMLINE_ACRONYM and ENDSTATION_ACRONYM environment variables.

jwlodek:
Because `sync-experiement` doesn't allow passing in the redis server url, the auto-deduction of the url needs to work for cases where location matches two hosts, like at XPD. Use the beamline/endstation acronym environment variables to do this.


RobertSchaffer1 (approved):

jwlodek:
@copilot resolve the merge conflicts in this pull request

PR #258, opened 2026-05-29
fixed warmup method in Xspress3HDF5Plugin

### LINKED ISSUE #257: Xspress3HDF5Plugin warmup() tries to set disabled PVs
The `warmup` method inherited from `HDF5Plugin` in `Xspress3HDF5Plugin` sets two PVs that are disabled in the community Xspress3 IOC, `self.parent.cam.image_mode` and `self.parent.cam.acquire_period`, which causes `warmup()` to fail.

dihenriksen:
Fixes https://github.com/NSLS2/nslsii/issues/257

### DIFF
--- a/nslsii/areadetector/xspress3.py
+++ b/nslsii/areadetector/xspress3.py
@@ -3,7 +3,7 @@ import logging
 import re
 import time as ttime
 
-from collections import deque
+from collections import deque, OrderedDict
 from pathlib import Path
 from uuid import uuid4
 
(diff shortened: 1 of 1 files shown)


jwlodek:
Since the method is identical to the upstream one other than the two signals, I wonder if we should change warmup upstream to take a dictionary of signals to values to set? Maybe not worth the effort, but it would certainly allow for less code duplication...

Or maybe have a property that stores the dict of signals to set on warmup.

dihenriksen:
> Since the method is identical to the upstream one other than the two signals, I wonder if we should change warmup upstream to take a dictionary of signals to values to set? Maybe not worth the effort, but it would certainly allow for less code duplication...
> 
> Or maybe have a property that stores the dict of signals to set on warmup.

I thought about that too. It might be the best and cleanest solution - it is easy for me to imagine other detectors having to tweak the signals for the HDF5Plugin as well. But since that change would be in ophyd I am worried about unintended downstream consequences and backwards compatibility. And I'm still only vaguely familiar with ophyd so I wasn't confident in making that change there.

dihenriksen:
@jwlodek I at least made in issue in ophyd: https://github.com/bluesky/ophyd/issues/1280

PR #263, opened 2026-07-18
Update packaging to pyproject; Fix test errors from ophyd-async changes

### LINKED ISSUE #206: Update `nslsii` to use `pyproject.toml`
We should use https://github.com/scientific-python/cookie (we use it already for [`hextools`](https://github.com/NSLS-II-HEX/hextools), [`srx-caproto-iocs`](https://github.com/NSLS-II-SRX/srx-caproto-iocs), etc.).

Add a `[sync-experiment]` install candidate, so we can do `pip install .[sync-experiment]` instead of having to track dependencies in multiple places.

(lifted from https://github.com/NSLS2/ansible/pull/2325#issuecomment-2447750489)

### LINKED ISSUE #261: Tests have been failing since June 4
Tests have been [failing since June 4](https://github.com/NSLS2/nslsii/actions?page=2).

This should be addressed in a separate PR.

_Originally posted by @padraic-shafer in https://github.com/NSLS2/nslsii/issues/260#issuecomment-4884265663_

thopkins32:
Assisted-by: opencode:gpt-5.5

# Summary

Switches the project packaging infrastructure to use more modern tooling like:
- pyproject.toml
- hatchling & hatch-vcs
- pixi for development

It also fixes some broken tests due to breaking API changes in newer releases of ophyd-async.

Most lines changed are from `pixi.lock`.

## Other decisions

- Tried to keep everything backward compatible
- Drops Python 3.10 support (due to ophyd-async dependency)
- Merges `pixi.toml` under `pyproject.toml` to get auto-use of `[dependency-groups]` as Pixi features
- Added `ruff` formatting and linting config, but disabled it for now, I think the right way to do this is file-by-file in separate PRs rather than all at once, to not disrupt active PRs from having to deal with so many merge conflicts. 

## Future work

- Rework pre-commit
- Remove flake8
- Rework GitHub workflows (remove Travis CI config)
- Add test coverage
- Split dependencies into extras

Closes #206 
Closes #261


thopkins32:
ophyd-async droppped Python 3.10 https://github.com/bluesky/ophyd-async/blob/2919708f3dfb392ec99521f3294a95d746b43310/pyproject.toml#L31

I can work around it by excluding those tests for Python 3.10, or if we can Python 3.10 support.

jwlodek:
I'm OK with dropping 3.10, if other s are too. Could we move the dev dependencies into a 'dependency group' instead of an optional extra? Also would be good to split up the base dependencies into extras, or consider if some can be removed. Conda now supports extra install candidates so we should update the conda package to use those too. Ideally the base install should only include the dependencies required for a minimal setup.

thopkins32:
> Conda now supports extra install candidates

This is news to me. I couldn't find much about this other than that it's coming soon.

Edit: It looks like the proposal was recently accepted. https://github.com/conda/ceps/blob/main/cep-0044.md

jwlodek:
Yeah it's very new I think the CEP was only accepted a month or two ago.

thopkins32:
The only way I got the dependency-group to work nicely with Pixi was to merge `pixi.toml` under `pyproject.toml` using `tool.pixi.` syntax.

thopkins32:
> Also would be good to split up the base dependencies into extras, or consider if some can be removed. Conda now supports extra install candidates so we should update the conda package to use those too. Ideally the base install should only include the dependencies required for a minimal setup.

@jwlodek I'm going to hold off on this for this PR, can be a follow-up item.

pyproject.toml
```diff
@@ -0,0 +1,156 @@
...
+  "ipython",
+  "ipywidgets",
+  "ldap3",
+  "matplotlib",
+  "msgpack>=1.0.0",
+  "msgpack-numpy",
+  "numpy",
+  "opencv-python",
+  "ophyd",
+  "ophyd-async",
```
sligara7: may need `ophyd-async[ca]`

pyproject.toml
```diff
@@ -0,0 +1,156 @@
...
+  "attrs>=17.4.0",
+  "build",
+  "coverage",
+  "doct",
+  "ipython",
+  "matplotlib",
+  "numpydoc",
+  "pims",
+  "pre-commit",
+  "pytest",
```
sligara7, line 64: can we also add `pytest-asyncio`?

pyproject.toml
thopkins32: Good call, thanks.

sligara7 (approved):
Thanks for working this.  I have a PR in also and was having issues because, as you noted, ophyd-async >0.19 doesn't support python 3.10 any longer.  So this should help there.

PR #264, opened 2026-07-21
Add Eurotherm3k ophyd-async device (HXM-1305)

sligara7:
Add nslsii.ophyd_async.devices.eurotherm3k: Eurotherm3kLoop (a Movable PID loop with a settle-aware set()) plus a two-loop Eurotherm3k container and its StrictEnums, mapping the nsls2.ioc_deploy eurotherm3k IOC role. Export the new classes from the devices package and fix the pre-existing broken absolute 'from rbd9103 import ...' there (-> relative). Add mock-backend unit tests including a legacy PV-surface equivalence check against the classic Eurotherm, and add pytest-asyncio to the dev requirements.

Assisted-by: copilot:claude-opus-4-8

### DIFF
--- a/nslsii/ophyd_async/devices/__init__.py
+++ b/nslsii/ophyd_async/devices/__init__.py
@@ -1,10 +1,17 @@
-from rbd9103 import (
+from .rbd9103 import (
     RBD9103,
     RBD9103Range,
     RBD9103Input,
     RBD9103SamplingMode,
     RBD9103InRangeState,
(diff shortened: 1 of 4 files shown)


thopkins32 (commented):
Did you consider using `StandardMovable` instead of `StandardReadable`? It could simplify the implementation: https://blueskyproject.io/ophyd-async/main/explanations/when-to-extend-movable.html#using-standardmovable

PR #266, opened 2026-08-25
Optionally store git metadata in RE md

josephhanrahan:
As part of an effort to have more provenance for [blop](https://github.com/bluesky/blop) agents, having the ability to link profile collection code back to an optimization run is important.

This PR adds an optional argument `profile_collection_dir` to `configure_base()` which when provided, stores the current ref, branch name, and dirty status to `RE.md`

### DIFF
--- a/nslsii/__init__.py
+++ b/nslsii/__init__.py
@@ -49,7 +49,7 @@ def configure_base(
     redis_ssl=False,
     redis_prefix="",
     redis_db: int = 0,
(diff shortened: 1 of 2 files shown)


jwlodek:
I like the idea, but could we use naming that isn't tied to `profile_collection`? Maybe `startup_dir`? It's possible (likely) we'll move away from `profile-collection`s in the future.

jwlodek (approved):

